PR #6, opened 2015-03-30
Add security code field to token request

sirn:
Add the missing `setSecurityCode` in token request and properly set `securityCodeCheck` in JSON parser.


robinclart:
:+1:

PR #15, opened 2016-10-04
Enable TLS 1.2 in Android prior to 5.0.0

pitiphong-p:
For more information please go to https://github.com/square/okhttp/issues/1934

### DIFF
--- a/sdk/src/main/java/co/omise/android/Client.java
+++ b/sdk/src/main/java/co/omise/android/Client.java
@@ -1,17 +1,27 @@
 package co.omise.android;
 
+import android.os.Build;
 import android.os.Handler;
 
 import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.util.Arrays;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
 
 import okhttp3.CertificatePinner;
 import okhttp3.Credentials;
 import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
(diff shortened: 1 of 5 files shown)


sdk/src/main/java/co/omise/android/Client.java
```diff
@@ -50,8 +60,35 @@ public void run() {
...
+                    TrustManagerFactory.getDefaultAlgorithm());
+            trustManagerFactory.init((KeyStore) null);
+            TrustManager[] trustManagers = trustManagerFactory.getTrustManagers();
+            if (trustManagers.length != 1 || !(trustManagers[0] instanceof X509TrustManager)) {
+                throw new IllegalStateException("Unexpected default trust managers:"
+                        + Arrays.toString(trustManagers));
+            }
+            return (X509TrustManager) trustManagers[0];
+        } catch (GeneralSecurityException e) {
+            throw new AssertionError(); // The system has no TLS. Just give up.
```
chakrit: why `AssertionError`?

sdk/src/main/java/co/omise/android/Client.java
```diff
@@ -50,8 +60,35 @@ public void run() {
...
+
     private OkHttpClient buildHttpClient(final String publicKey) {
-        return new OkHttpClient.Builder()
+        OkHttpClient.Builder builder = new OkHttpClient.Builder();
+        X509TrustManager trustManager = systemDefaultTrustManager();
+        if (Build.VERSION.SDK_INT < 21) {
+            try {
+                builder.sslSocketFactory(new TLSSocketFactory(), trustManager);
+            } catch (Exception ex) {
+                ex.printStackTrace();
```
chakrit: Can we also emit a warning here so the SDK user is aware of what's happening?

chakrit: Hmm we don't need to even call `systemDefaultTrustmanager()` under `SDK_INT >= 21` right? So let's skip all these shenanigans on platforms that already works correctly.

PR #23, opened 2017-07-17
Built-in Authorizing Payment scene

pitiphong-p:
Since the Authorizing Payment is a common operation when using Omise service now. I think it would be great if we have the built-in way to support the Authorizing Payment in the SDK.

This PR introduces the `AuthorizingPaymentActivity` scene which replaces the old `Verify3DSActivity`. This new OmiseAuthorizingPaymentViewController requires the expected return URL patterns in order to make it work properly. The merchants/library-users need to specify their return URL patterns to this scene.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -37,7 +37,7 @@
     <ConfirmationsSetting value="0" id="Add" />
     <ConfirmationsSetting value="0" id="Remove" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" assert-keyword="true" jdk-15="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" default="true" assert-keyword="true" jdk-15="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/build/classes" />
   </component>
   <component name="ProjectType">
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -37,7 +37,7 @@
     <ConfirmationsSetting value="0" id="Add" />
     <ConfirmationsSetting value="0" id="Remove" />
   </component>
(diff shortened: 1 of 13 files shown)


chakrit:
Looks like we need to bump minor version with this.

pitiphong-p:
Yeah, will do that after this PR got merged

.idea/misc.xml
```diff
@@ -37,7 +37,7 @@
     <ConfirmationsSetting value="0" id="Add" />
     <ConfirmationsSetting value="0" id="Remove" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" default="true" assert-keyword="true" jdk-15="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" assert-keyword="true" jdk-15="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
chakrit: @pitiphong-p 🤔 intentional?

I think we need to stick to Java 7 na.

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        expectedReturnURLPatterns = new Uri[returnURLPatternList.size()];
+        expectedReturnURLPatterns = returnURLPatternList.toArray(expectedReturnURLPatterns);
+
+        if (authorizedURL != null && authorizedURL.toString() != null) {
+            webView.loadUrl(authorizedURL.toString());
+        }
+
+        webView.setWebViewClient(new WebViewClient() {
```
chakrit: @pitiphong-p Can we split this into an inner class inside this class? I think it'll be more readable on its own unit.

.idea/misc.xml
pitiphong-p: Nope, OK will try to find how to revert that

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.java
pitiphong-p: Ummm 🤔, sounds interesting. OK will try to do that

pitiphong-p:
Updated with the inner class `AuthorizingPaymentURLVerifier`

Issue #27, opened 2017-08-08
Samsung device space bar not added in CreditCardEditText


TheEnsis:
I tested and it worked now for sony and asus device

pitiphong-p:
Could you provide the device information? Model and Android version

TheEnsis:
FYI : I added it it inside TextInputLayout

Devices I tested are samsung galaxy j5 version 6.0.1 and samsung note 5 version 7.0
Devices that work fine is sony XZs version 7.1 and asus zenphone version 6.0.1

chakrit:
@nuxZERO please check if you can krub 🙇

WeRockStar:
Huawei has same issue.

nuxzero:
@WeRockStar What the model and Android version?

WeRockStar:
@nuxzero Huawei P10 Plus and Android 7.0. I add EditText inside TextInputLayout (Floating Label)

nuxzero:
@WeRockStar we already fixed about spacing on `CreditCardEditText `. Please checkout our new version `v2.6.2`.

Please don't hesitate to contract us If you has any problem.

WeRockStar:
@nuxzero Thank you for support 🤘

PR #30, opened 2017-10-30
Fix credit card number format.

nuxzero:
This PR fixed issue #27. Because of `OnKeyListener` will fired only hardware keyboard and not working with alternative keyboards. To solve format on `CreditCardEditText`, use `TextWatcher` instead of `OnKeyListener` to format credit card number.

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -81,8 +81,6 @@ uploadArchives {
 dependencies {
     compile 'joda-time:joda-time:2.9.2'
     compile 'com.squareup.okhttp3:okhttp:3.4.1'
-    compile 'com.android.support:appcompat-v7:26.1.0'
-    compile 'com.android.support.constraint:constraint-layout:1.0.2'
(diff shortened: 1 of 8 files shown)


nuxzero:
@pitiphong-p @chakrit please review krub.

sdk/build.gradle
```diff
@@ -80,6 +81,13 @@ uploadArchives {
 dependencies {
     compile 'joda-time:joda-time:2.9.2'
     compile 'com.squareup.okhttp3:okhttp:3.4.1'
+    compile 'com.android.support:appcompat-v7:26.1.0'
```
pitiphong-p: Can we do this without support library? We try to avoid that for the compatibility issue with our client.

nuxzero: This one using for testing with `AppCompatActivity`. Btw I can change to normal `Activity` and delete support library.

nuxzero: @pitiphong-p please review gain krub.

PR #31, opened 2017-10-31
Update user agent.

nuxzero:
Update `User-Agent` to `OmiseAndroid/2.6.1 Android/26 Model/Nexus 5X` format.


nuxzero:
@pitiphong-p Yes

PR #35, opened 2018-01-09
Handle authorize uri with external app

nuxzero:
Supported authorize URI with external app, for example Alipay app. `AuthrizingPaymentActivity` automatically open external app when authorize URI was not `https, http, about` scheme. When user confirmed payment from external app, then will pass result back to `onActivityResult()` in merchant app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,9 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 }
 ```
 
+#### Authorizing Payment via an external app
+
+Some request method allow user to authorize with external app, for sample Alipay. When user would like to authorize their payment with external app, `AuthorizingPaymentActivity` will automatically open external app by default. However merchant developers handle the `Intent` callback by themselves.
 
 ## ProGuard Rules
 
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
 #### Authorizing Payment via an external app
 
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.java
```diff
@@ -103,12 +112,24 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
                 }
             }
         });
     }
-}
 
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        if (requestCode == REQUEST_EXTERNAL_CODE && resultCode == RESULT_OK) {
```
pitiphong-p: Should we verify the result in the returned Intent too?

nuxzero: The result is depend on external app. By default can check only `RESULT_OK` or `RESULT_CANCELED`. So, merchant developer must handle `Intent` result back by themselves na krub.

README.md
```diff
@@ -240,6 +240,9 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 }
 ```
 
+#### Authorizing Payment via an external app
+
+Some request method allow user to authorize with external app, for sample Alipay. When user would like to authorize their payment with external app, `AuthorizingPaymentActivity` will automatically open external app by default. However merchant developers handle the `Intent` callback by themselves.
```
chakrit: @nuxzero Please run this through one of the native speaker in office or the thesaurus/grammar checker krub.

PR #47, opened 2018-09-21
Drop support TLS 1.1 and add doc to support Android older version 5.0

nuxzero:


README.md
```diff
@@ -257,6 +257,34 @@ If you enable ProGuard, then add this rules in your ProGuard file.
...
+        e.printStackTrace();
+    } catch (GooglePlayServicesNotAvailableException e) {
+        e.printStackTrace();
+    }
+}
+```
+
+> **Note:** This will support only the devices that have Google Play Services installed on the device.
+
+
```
Namiii: There are a few things that can be improved here so it's just easier for me to copy the whole thing here and add the corrections 👍 

## Support TLS 1.2 for devices with OS older than Android 5.0  

To follow PCI compliance we will support TLS 1.2 or newer on our vault. The applications that are using Omise Android SDK and are running on OS versions older than Android 5.0 will stop working. This is because Android API older than version 20 have TLS 1.2 disabled by default for openssl security reason. To keep Omise Android SDK working on Android devices with versions older than 5.0, developers need to use Google Play Services to install a new patch of SSL Security Provider.
 
If use already use one of Google Play Services libraries e.g. **GCM**, **Analytics** you are already able to use `ProviderInstaller` from them. But if not developers need to add one of the Google Play Services libraries to their project. 
 ```gradle
compile 'com.google.android.gms:play-services-analytics:16.0.3'
```
 After adding Google Play Services to the project, developers need to install Google Play Services in the Application class or in the Activity before they start to call Omise Android SDK. Developers can install that by following this snippet of code.
 ```java
 if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
    try {
        ProviderInstaller.installIfNeeded(this);
    } catch (GooglePlayServicesRepairableException e) {
        e.printStackTrace();
    } catch (GooglePlayServicesNotAvailableException e) {
        e.printStackTrace();
    }
}
```
 > **Note:** This will only support the devices that have Google Play Services installed on them.

nuxzero: @Namiii Thank you. I will wait for other feedback then will combine them.

pitiphong-p: Here's my version of the note

## Note on TLS 1.2
PCI-DSS standard requires the service to communicate in TLS 1.2 or higher using strong encryption suite. This means that every clients must connect to Omise service with TLS 1.2 or higher using some valid strong encryption suites. However TLS 1.2 in Android is vary depends on the Android OS

#### Android API 21 or higher
You can use our SDK without any change. The SDK already fully supports communication to the Omise Service using TLS 1.2 with the properly encryption suite.
#### Android API 16 to API 20 with Google Play Services
These Android API versions don't support the proper encryption suites out of the box. However `Google Play Service` has at the `ProviderInstaller` API to add support for the proper encryption suite. You may already use the Google Play Service in your app already; Google Play Service includes many common libraries used in many Android apps including GCM, Analytics and more. You can add support for the proper encryption suite with Google Play Service by doing the following steps
1. Add Google Play Service Analytics to your app grade setting
```gradle
 compile 'com.google.android.gms:play-services-analytics:16.0.3'
 ```
2. Asks the ProviderInstaller to install the encryption suites with the follow code
```java
 if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
     try {
         ProviderInstaller.installIfNeeded(this);
     } catch (GooglePlayServicesRepairableException e) {
         e.printStackTrace();
     } catch (GooglePlayServicesNotAvailableException e) {
         e.printStackTrace();
     }
 }
 ```
> **Note:** Google Play Service may not available on every brands or models. Please be concerned about the Google Play Service compatibility
#### API 15 or lower
These Android API levels could not add support for the proper encryption suites easily which means that the Android devices running those OS versions may not be able to use to connect to many services on the internet. We recommend you to drop support for those API levels

**Comment by nuxzero** [comment]
It ambiguous to say Android API 20. Exactly Android 4.4 but for Android wear.

**Comment by nuxzero** [comment]
BTW TLS 1.2 enable since Android API 20. I will change a bit of API version.

File: README.md [hunk 2]
```diff
@@ -257,10 +257,46 @@ If you enable ProGuard, then add this rules in your ProGuard file.
...
+     } catch (GooglePlayServicesNotAvailableException e) {
+         e.printStackTrace();
+     }
+ }
+ ```
+ 
+> **Note:** Google Play Service may not available on every brands or models. Please be concerned about the Google Play Service compatibility
+
+#### API 15 or lower
+These Android API levels could not add support for the proper encryption suites easily which means that the Android devices running those OS versions may not be able to use to connect to many services on the internet. We recommend you to drop support for those API levels
```
pitiphong-p: These Android API levels could not ~~add~~ support for the proper encryption suites easily which means that the Android devices running those OS versions may not be able to use to connect to *Omise service and also* many services on the internet. We recommend you to drop support for those API levels

Please remove the word `add` in the first sentence and add a phase that I emphasize in the text above

nuxzero:
@pitiphong-p ok, I will update.

README.md
```diff
@@ -257,10 +257,46 @@ If you enable ProGuard, then add this rules in your ProGuard file.
 ```
 
 
+## Note on TLS 1.2
+PCI-DSS standard requires the service to communicate in *TLS 1.2* or higher using strong encryption suite. This means that every clients must connect to Omise service with those valid suites. However TLS 1.2 support in Android is vary depends on the Android OS.
+
```
pitiphong-p: Please add `Please follow the following instruction to add support for TLS 1.2 in your app` before the new heading

Sorry I missed again. However you can merge this PR after you add it here :)

nuxzero: @pitiphong-p Do you mean the end of this paragraph?

pitiphong-p: Correct. Right where I mark for this comment

nuxzero:
@pitiphong-p added krub.

Closing this PR.

PR #49, opened 2018-09-25
CreditCardActivity setResult change

Namiii:
Small PR where we've changed the behavior of `CreditCardActivity` to setResult using the default codes provided by the SDK instead of custom codes. 

Task -> https://phabricator.omise.co/T7791


nuxzero:
@Namiii Could update in README file too?

Namiii:
@nuxzero done 👍

PR #57, opened 2019-07-04
Refactor request flow

Namiii:
**TASK**=>[T13471](https://phabricator.omise.co/T13471)

**SUMMARY**
In this PR, we have created a general `Request` that can be passed through the `Client` in order to enable us to have a number of different API calls in the SDK, instead of a hardcoded `TokenRequest`. A base `RequestBuilder` class was also made to be used to create different Requests. Along with this, we have also made changes to the Request callback to return a `Model` instead of a `Token`. In this refactor, a number of classes were also converted to Kotlin.

**QA**
The Token creation process should work as it did before.

**FURTHER WORK**
There are a couple of areas that need to be improved but are segmented off as their own tasks that need to be done after this PR is merged.
- Integrating a Parsing library [T13595](https://phabricator.omise.co/T13595)
- Adding a test flow for Requests [T13596](https://phabricator.omise.co/T13596)


sdk/src/main/java/co/omise/android/api/Client.kt
```diff
@@ -0,0 +1,79 @@
...
+            chain.proceed(chain.request()
+                    .newBuilder()
+                    .addHeader("User-Agent", buildUserAgent())
+                    .addHeader("Authorization", Credentials.basic(publicKey, "x"))
+                    .build())
+        }
+    }
+
+    private fun buildUserAgent(): String {
+        return "OmiseAndroid/" + BuildConfig.VERSION_NAME +
```
pitiphong-p: Should we have a settable app name in the User Agent here?

Namiii: Hmmm as in have `OmiseAndroid/` as config too? Think we can but do you have any use case in mind for it? For dev and release maybe?

sdk/src/main/java/co/omise/android/api/Invocation.kt
```diff
@@ -0,0 +1,78 @@
...
+    private fun processCall(call: TypedCall) {
+        val response = call.execute()
+        if (response.body() == null) {
+            didFail(IOException("HTTP response have no body."))
+            return
+        }
+
+        val rawJson = response.body().string()
+        val model = parseModelFromJson(rawJson, call)
+        if (response.code() in 200..299 && model != null) {
```
pitiphong-p: I think we should have a special case for handling 300..399 status codes? We shouldn't allow redirection and have a specific case.

Namiii: So, a custom exception?

pitiphong-p: Correct

Namiii: What would we call this exception btw? :D

sdk/src/main/java/co/omise/android/api/RequestBuilder.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Abstract method that needs to be implement by all children of this class to provide response type
+     *
+     * @return Class type of response
+     */
+    protected abstract fun type(): Class<T>
+
+    companion object {
```
nuxzero: Shall we use **sealed class** for request method?

sdk/src/main/java/co/omise/android/api/Invocation.kt
```diff
@@ -0,0 +1,78 @@
...
+        private val replyHandler: Handler,
+        private val httpClient: OkHttpClient,
+        private val request: co.omise.android.api.Request<T>,
+        private val listener: RequestListener<T>
+) {
+
+    fun invoke() {
+        val call = httpClient.newTypedCall(
+                Request.Builder()
+                        .method(request.method,
```
nuxzero: Should we make it same line?

Namiii: Which ones?

sdk/src/main/java/co/omise/android/api/RequestBuilder.kt
Namiii: I think we had this discussion for Omise Java as well :) The conclusion was that just having it as String is enough and there is no need to overcomplicate things.

nuxzero: Yeah, that mean we need to take this risky. BTW, it's OK.

sdk/src/main/java/co/omise/android/api/Invocation.kt
nuxzero: `method()`

Namiii: Hmmm, alright 👌

Namiii: @nuxzero how about this? 

`
        val call = httpClient.newTypedCall(Request.Builder()
                .method(request.method,
                        request.payload)
                .url(request.url)
                .build(),
                request.responseType)
`

Namiii: @nuxzero How about this? 

E5C70AAD-F1E5-401D-B55F-D5075FD15CE5

nuxzero: No no, I mean 
```
val call = httpClient.newTypedCall(
                Request.Builder()
                        .method(request.method, request.payload)
                        .url(request.url)
                        .build(),
                request.responseType)
```

nuxzero: BTW, approved. 👍

Namiii: Ahhhh okk, now I get it 👌

PR #58, opened 2019-07-08
Adopt AndroidX to AuthorizingPaymentActivity

nuxzero:
### OBJECTIVE

We designed to update the `AuthorizingPaymentActivity` to use `AppCompatActivity`(a compatibility version of `Activity`) because of the trend of creating the application use **AndroidX(Support Library)**.

### DESCRIPTION OF CODE CHANGE

On the `AuthorizingPaymentActivity` has changed to inherit from `AppCompatActivity`. Also, convert the class to **Kotlin** and created the test. 

### QA

The code change does not affect to the user side. All unit test should pass.


sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -72,7 +70,7 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
     }
 
     override fun onBackPressed() {
-        setResult(Activity.RESULT_CANCELED)
+        setResult(RESULT_CANCELED)
```
igroomgrim: 👍

PR #59, opened 2019-07-08
Capabilities api request

Namiii:
**SUMMARY** 
In this PR we have added `Capabilities API` as a request that can be used by the user to retrieve their account's capabilities. This required reworking the `Client` setup.

**QA**
Check that the Capabilities request returns a correct model as response.


nuxzero:
Why you need to use `skey`?

Namiii:
@nuxzero for capabilities API, otherwise you get auth error

nuxzero:
I don't think it needs `skey`. What the error you got?

Namiii:
Authentication error

nuxzero:
From the doc looks like it doesn't required the `skey` https://www.omise.co/capability-api.

Also, I was tested via Postman it work fine. 🤔

app/src/main/java/co/omise/android/example/CheckoutActivity.java
```diff
@@ -14,7 +14,8 @@
 import co.omise.android.ui.CreditCardActivity;
 
 public class CheckoutActivity extends BaseActivity implements View.OnClickListener {
-    public static final String OMISE_PKEY = "pkey_test_52d6po3fvio2w6tefpb";
+    public static final String OMISE_PKEY = "pkey_test_5cotumuix3r6g414148";
+    public static final String OMISE_SKEY = null;
```
pitiphong-p: Asking for an skey is a hard NO for me

Namiii:
@nuxzero I don't know about that, what I know is what I'm getting through the API here,


This is the response you get if you only use `public key`
2B607B25-2DBF-4B64-A758-6D8C1F870D10
C7D4BC85-663D-4D90-A884-6F8C9CC9E418


This is what you get if you only use `secret key` 
9CC94421-3D00-4270-AD24-98D561B00FBB
7173DB4D-700C-41DE-B715-904349E4721E

This is the other way around if you are using `vault`

Namiii:
@nuxzero @pitiphong-p I'll dig around and see what I can get out of it without `skey`, but that's how it's behaving

pitiphong-p:
@Namiii Maybe because you fix the key in the Endpoint? https://github.com/omise/omise-android/blob/d741df8f0c31d5c120d5bc891d72d13af048f93c/sdk/src/main/java/co/omise/android/api/EndPoint.kt#L83

Namiii:
@pitiphong-p Yeah I was just changing that

Namiii:
@nuxzero @pitiphong-p changes made

sdk/src/main/java/co/omise/android/api/Client.kt
```diff
@@ -2,28 +2,44 @@ package co.omise.android.api
...
 
+
 /**
  * Client is the main entry point to the SDK and it needs to be supplied with a public Key. You can use the Client to send a [Request].
  *
  * @param publicKey The key with the `pkey_` prefix.
  * @see Request
  */
-class Client(private val publicKey: String) {
+class Client(publicKey: String?)
```
pitiphong-p: Why is the publicKey a Nullable? I think publicKey is a required value

sdk/src/main/java/co/omise/android/models/PaymentMethod.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    override fun writeToParcel(parcel: Parcel, flags: Int) {
+        super.writeToParcel(parcel, flags)
+        parcel.writeString(name)
+        parcel.writeStringList(currencies)
+        parcel.writeStringList(cardBrands)
+    }
+
+    override fun describeContents(): Int {
+        return 0
```
pitiphong-p: What is this property for?

igroomgrim:
👍

sdk/src/main/java/co/omise/android/models/Capability.kt
```diff
@@ -0,0 +1,62 @@
...
+    var banks: List<String>? = null
+    @JvmField
+    var paymentMethods: List<PaymentMethod>? = null
+    @JvmField
+    var zeroInterestInstallments: Boolean = false
+
+
+    constructor(rawJson: String) : this(JSONObject(rawJson))
+
+    private constructor(json: JSONObject) : super(json) {
```
nuxzero: I think we should use any tools (i.e. Jackson, Gson) to help us parsing Json. However, we should another task to do the serializer.

sdk/src/main/java/co/omise/android/models/Capability.kt
```diff
@@ -0,0 +1,62 @@
...
+     * The {@link RequestBuilder} class for retrieving account Capabilities.
+     */
+    class GetCapabilitiesRequestBuilder : RequestBuilder<Capability>() {
+
+        override fun path(): HttpUrl = buildUrl(Endpoint.API, "capability")
+
+        override fun type(): Class<Capability> = Capability::class.java
+    }
+
+    override fun writeToParcel(parcel: Parcel, flags: Int) {
```
nuxzero: What about using **Kotlin Serializable** instead of manually build Pracel?

sdk/src/main/java/co/omise/android/models/PaymentMethod.kt
```diff
@@ -0,0 +1,43 @@
+package co.omise.android.models
+
+import android.os.Parcel
+import android.os.Parcelable
+import org.json.JSONObject
+
+class PaymentMethod : Model {
```
nuxzero: Data class?

sdk/src/main/java/co/omise/android/api/EndPoint.kt
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * The authentication key to use. The key should be taken from the given [Config] object.
+     * Either [Config.publicKey] or [Config.secretKey] should be returned.
+     *
+     * @param config A [Config] instance.
+     * @return A [String] containing the authentication key.
+     */
+    abstract fun authenticationKey(config: Config): String?
```
nuxzero: I don't think it should return nullable data.

sdk/src/main/java/co/omise/android/api/Config.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Config class bundles configuration values supplied to the [Client] constructor.
+ *
+ * @param apiVersion The API version to use.
+ * @param publicKey  The key with the `pkey_` prefix.
+ *
+ * @see Client
+ */
+class Config(
+        private val apiVersion: String?,
+        private val publicKey: String?
```
nuxzero: I think the constructor should not nullable.

nuxzero:
How can I test the capabilities request?

app/src/main/java/co/omise/android/example/CheckoutActivity.java
```diff
@@ -14,7 +14,7 @@
 import co.omise.android.ui.CreditCardActivity;
 
 public class CheckoutActivity extends BaseActivity implements View.OnClickListener {
-    public static final String OMISE_PKEY = "pkey_test_52d6po3fvio2w6tefpb";
+    public static final String OMISE_PKEY = "pkey_test_5cotumuix3r6g414148";
```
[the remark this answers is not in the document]
igroomgrim: Don't forget to revoke it 😄

sdk/src/main/java/co/omise/android/api/Client.kt
Namiii: Oh yeah, leftout `?` from when I was using both, should be removed.

sdk/src/main/java/co/omise/android/models/PaymentMethod.kt
Namiii: Just carries on from parent `Model` class that has `Parcelable` implemented in it.

app/src/main/java/co/omise/android/example/CheckoutActivity.java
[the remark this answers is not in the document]
Namiii: Oh shit :scream:

app/src/main/java/co/omise/android/example/MainActivity.java
```diff
@@ -51,7 +51,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
         if (item.getItemId() == R.id.menu_authorizing_payment_action) {
             Intent intent = new Intent(this, AuthorizingPaymentActivity.class);
             intent.putExtra(EXTRA_AUTHORIZED_URLSTRING, "https://pay.omise.co/offsites/");
-            intent.putExtra(EXTRA_EXPECTED_RETURN_URLSTRING_PATTERNS, new String[] {"http://www.example.com"} );
+            intent.putExtra(EXTRA_EXPECTED_RETURN_URLSTRING_PATTERNS, new String[]{"http://www.example.com/orders"});
```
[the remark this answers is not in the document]
Namiii: Sure

sdk/src/main/java/co/omise/android/api/EndPoint.kt
```diff
@@ -0,0 +1,105 @@
...
+     * pin no certificate.
+     */
+    fun certificateHash(): String {
+        return "sha256/maqNsxEnwszR+xCmoGUiV636PvSM5zvBIBuupBn9AB8="
+    }
+
+    /**
+     * The authentication key to use. The key should be taken from the given [Config] object.
+     * Either [Config.publicKey] or [Config.secretKey] should be returned.
+     *
```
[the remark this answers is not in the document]
Namiii: 👍

sdk/src/main/java/co/omise/android/models/Capability.kt
```diff
@@ -0,0 +1,62 @@
...
+import co.omise.android.api.RequestBuilder
+import okhttp3.HttpUrl
+import org.json.JSONObject
+
+/**
+ * Represents Capabilities object and contains its {@link RequestBuilder}.
+ *
+ * @see <a href="https://www.omise.co/capability-api">Capabilities API</a>
+ */
+class Capability : Model {
```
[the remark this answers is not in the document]
Namiii: Can a `data class` extend from a normal java class? And do we gain anything here from using a data class?

sdk/src/main/java/co/omise/android/models/Capability.kt
Namiii: As you also know, this is not my preferred method of parsing and this was already in place since 2016. This [task](https://phabricator.omise.co/T13595) is already in place for us to discuss and choose a suitable serializer library 👍

sdk/src/main/java/co/omise/android/models/Capability.kt
Namiii: We can also do this as part of the Serialization task. But does if we use `Kotlin Serializable`, do Java people also get serializability?

sdk/src/main/java/co/omise/android/models/PaymentMethod.kt
Namiii: Same question as the last one from me

sdk/src/main/java/co/omise/android/api/EndPoint.kt
Namiii: Yeah leftover from having both keys, will change 👍

sdk/src/main/java/co/omise/android/api/Config.kt
Namiii: Same comment as above.

Namiii:
@nuxzero only manually by writing code. We will add request tests as part of its own task to make this better.

sdk/src/main/java/co/omise/android/models/Capability.kt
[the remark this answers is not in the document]
nuxzero: Yeah, it can but I think you might need refactor `Model` before change. So, the gain as you know ability of **data class** and I think it more suitable at here.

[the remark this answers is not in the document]
nuxzero: BTW, we can keep it like this for now and do it in another PR.

sdk/src/main/java/co/omise/android/models/Capability.kt
nuxzero: 🆗

sdk/src/main/java/co/omise/android/models/Capability.kt
nuxzero: Yes, it's support.

sdk/src/main/java/co/omise/android/api/EndPoint.kt
nuxzero: 👌

sdk/src/main/java/co/omise/android/api/Config.kt
nuxzero: 👌

sdk/src/main/java/co/omise/android/models/Capability.kt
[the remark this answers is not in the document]
Namiii: Okk if that works, it should be fine. Maybe I'll convert all model classes to Kotlin while changing Serialization and also make them data classes. 👍

sdk/src/main/java/co/omise/android/models/Capability.kt
Namiii: okk then we'll implement it on all models in the Serialization PR 👍

sdk/src/main/java/co/omise/android/api/Config.kt
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Config class bundles configuration values supplied to the [Client] constructor.
+ *
+ * @param apiVersion The API version to use.
+ * @param publicKey  The key with the `pkey_` prefix.
+ *
+ * @see Client
+ */
+class Config(
+        private val apiVersion: String?,
```
pitiphong-p: I think we should pin to the latest API version here. Since we need the schema to be persistent to parse the result.

nuxzero: Same ☝️

sdk/src/main/java/co/omise/android/models/PaymentMethod.kt
```diff
@@ -0,0 +1,43 @@
...
+    var name: String? = null
+    var currencies: List<String>? = null
```
pitiphong-p: Is it possible to have this property and `name` to be NonNull and those the other twos below if it's possible.

sdk/src/main/java/co/omise/android/models/PaymentMethod.kt
pitiphong-p: This fix 0 value looks like a code smell to me. Do you know what is this property used for? So that we can know how to properly implement it.

sdk/src/main/java/co/omise/android/api/Config.kt
Namiii: The latest API version is pinned is being pinned through `EndPoint`, is that not ok?

sdk/src/main/java/co/omise/android/models/PaymentMethod.kt
Namiii: Unfortunately, not at this moment because then we'll have to manually handle null cases, else it'll crash. Once we put in a serialization library like Jackson, then we can easily make such changes 👍

sdk/src/main/java/co/omise/android/api/Config.kt
nuxzero: Since we using fixed API version I don't think it necessary to have `apiVersion`. 🤔

Namiii: This just gives up flexibility if needed in the future, and it's pinning the version anyway so what we want is achieved. So I don't think it's that necessary to make changes here.

pitiphong-p: I understand your point. How's about demote the Config to internal access level and set the `apiVersion` to be the predefined hard code value? Also make it NonNull as well

sdk/src/main/java/co/omise/android/models/PaymentMethod.kt
pitiphong-p: I see. OK then

sdk/src/main/java/co/omise/android/api/Config.kt
Namiii: Hmmm that's a good idea, something like this `private val apiVersion: String = "xxxxxx"`?

pitiphong-p: Yup :)

Namiii: Ok sorted 👌 🙏

nuxzero: 👌

sdk/src/main/java/co/omise/android/api/Configurer.kt
```diff
@@ -0,0 +1,49 @@
...
+            val apiVersion = config.apiVersion()
+            val endpoint = Endpoint.allEndpointsByHost[request.url().host()]
+                    ?: throw UnsupportedOperationException("unknown endpoint: " + request.url().host())
+
+            val key = endpoint.authenticationKey(config)
+            var builder = request.newBuilder()
+                    .addHeader("User-Agent", config.userAgent())
+                    .addHeader("Authorization", Credentials.basic(key, "x"))
+
+            if (apiVersion != null && !apiVersion.isEmpty()) {
```
pitiphong-p: Do we still need this check?

Namiii: Hmmm actually no, I think I need to the change the type to not nullable in config 👍

PR #64, opened 2019-07-30
Jackson integration

Namiii:
**TASK**=>T13595

**SUMMARY**
In this PR we have used the `Jackson` library to be able to parse the API responses in a better and more clean way than manually passing `JSON` strings to each Model. 

**QA**
All the requests should works as they did previously.

**FURTHER WORK**
Removing `Parcelable` and replacing it with a more Kotlin compatible component is going to be done as its own task. Task=>T13872


sdk/src/main/java/co/omise/android/models/APIError.kt
```diff
@@ -0,0 +1,18 @@
+package co.omise.android.models
+
+import org.joda.time.DateTime
+
+/**
+ * Represents Error object.
+ *
+ * @see <a href="https://www.omise.co/errors-api">Errors API</a>
+ */
+class APIError(
```
nuxzero: Is it possible to use `data class` here?

sdk/src/main/java/co/omise/android/models/APIError.kt
```diff
@@ -0,0 +1,18 @@
...
+        @JvmField
```
nuxzero: If I understand correctly, `@JvmField` will not generate getter/setter. 

Why do you think we should add `@JvmField`?

sdk/src/main/java/co/omise/android/models/ModelTypeResolver.kt
```diff
@@ -0,0 +1,48 @@
+package co.omise.android.models
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import com.fasterxml.jackson.databind.DatabindContext
+import com.fasterxml.jackson.databind.JavaType
+import com.fasterxml.jackson.databind.jsontype.impl.TypeIdResolverBase
+import java.util.*
```
nuxzero: Could you use single import instead wildcard import?

sdk/src/main/java/co/omise/android/models/Serializer.kt
```diff
@@ -0,0 +1,102 @@
...
+
+class Serializer {
+    /**
+     * Returns the pre-configured [ObjectMapper] used for
+     * serializing and deserializing Omise API objects.
+     *
+     * @return An [ObjectMapper] instance.
+     */
+    var objectMapper: ObjectMapper
+    /**
```
nuxzero: Could you add new line between fields?

sdk/src/test/java/co/omise/android/ConfigTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class ConfigTest {
+    private static final String MODEL = Build.MODEL;
+    private static final String PKG_VERSION = BuildConfig.VERSION_NAME;
+    static final String API_VERSION = "new-shiny-version";
+    private static final int ANDROID_VERSION = Build.VERSION.SDK_INT;
+    static final String PKEY = "pkey_test_123";
+
+    @Test
+    public void testConfig() {
```
nuxzero: I think the test naming not quite right. Could follow the convention name?

sdk/src/test/java/co/omise/android/ConfigTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import android.os.Build;
+
+import org.junit.Test;
+
+import co.omise.android.api.Config;
+
+import static org.junit.Assert.assertEquals;
+
+public class ConfigTest {
```
nuxzero: Do you have any plan to convert it to Kotlin?

Namiii: Yeah I think with the testing task, we can rework and convert everything. Didn't want to do too much in this task.

sdk/src/main/java/co/omise/android/models/Serializer.kt
Namiii: 👍

sdk/src/main/java/co/omise/android/models/ModelTypeResolver.kt
Namiii: 👍

sdk/src/main/java/co/omise/android/models/APIError.kt
Namiii: Can't remember why I didn't do this, will add it

sdk/src/test/java/co/omise/android/ConfigTest.java
Namiii: Done 👍

sdk/src/main/java/co/omise/android/models/APIError.kt
Namiii: Same as the previous comment. One of the problems of going on holiday in the middle of a task is that you can't remember the logic behind certain things 🤦‍♂ Need to think on this one

nuxzero: I thought ask asked me different thing.😂

From what I see now. If we call the field from Java it'll call `error.code` instead of Java convention `error.getCode()`.

Namiii: Yeah I get that, I just can't remember why I've made this decision, for all our data classes too 🤔

Namiii:
@nuxzero changes mades, please have another look 🙏

PR #67, opened 2019-08-02
Sources api request

Namiii:
**TASK**=> T13472

**SUMMARY**
In this PR we have added Source API as a request that can be used by the user to create a Source.

**QA**
Check that the Source request returns a correct model as response.

**FURTHER WORK**
There are a couple of issues here that needed to be done as part of their own tasks:

- Serialization test needs to be fixed - T14297
- Change Token RequestBuilder payload creation - T14296


sdk/src/main/java/co/omise/android/models/Source.kt
```diff
@@ -0,0 +1,176 @@
...
+ * @see [Sources API](https://www.omise.co/sources-api)
+ */
+data class Source(
+        var type: SourceType? = null,
+        var flow: FlowType? = null,
+        var amount: Long = 0,
+        var currency: String? = null,
+        var barcode: String? = null,
+        var references: References? = null,
+        @field:JsonProperty("store_id")
```
pitiphong-p: Not sure if this is a good idea to have properties of every possible source type in one Source class. How’s about subclassing?

sdk/src/test/resources/data/objects/source_object.json
```diff
@@ -0,0 +1,19 @@
+{
```
pitiphong-p: I think we should have fixtures files for each source types

sdk/src/main/java/co/omise/android/models/Serializer.kt
```diff
@@ -98,6 +101,19 @@ class Serializer {
...
     }
 
+    /**
+     * Serializes the given [RequestBuilder] object to the provided output stream.
+     *
+     * @param outputStream The [OutputStream] to serialize the parameter into.
+     * @param builder      The [RequestBuilder] to serialize.
+     * @param <T>          The type of the parameter object to serialize.
+     * @throws IOException on general I/O error.
+    </T> */
```
nuxzero: Seems to be AS add `</T>` for the `@param <T>`. Could you delete it?

sdk/src/main/java/co/omise/android/models/SourceType.kt
```diff
@@ -0,0 +1,40 @@
...
+    object BillPaymentTescoLotus : SourceType("bill_payment_tesco_lotus")
+    object BarcodeAlipay : SourceType("barcode_alipay")
+    object Econtext : SourceType("econtext")
+    object TrueMoney : SourceType("truemoney")
+    object InstBankingBay : SourceType("installment_bay")
+    object InstFirstChoice : SourceType("installment_first_choice")
+    object InstBbl : SourceType("installment_bbl")
+    object InstKtc : SourceType("installment_ktc")
+    object InstKBank : SourceType("installment_kbank")
+    object Unknown : SourceType(null)
```
nuxzero: Is it possible to pass the raw value through the `Unknown` type?

sdk/src/main/java/co/omise/android/models/Serializer.kt
```diff
@@ -98,6 +101,19 @@ class Serializer {
...
+    @Throws(IOException::class)
+    fun <T : RequestBuilder<*>> serializeRequestBuilder(outputStream: OutputStream, builder: T) {
```
nuxzero: Shall we use `Model` as specific generic type?

sdk/src/main/java/co/omise/android/models/Serializer.kt
Namiii: Oh yeah, will do 👌

sdk/src/main/java/co/omise/android/models/SourceType.kt
Namiii: I don't think there is a way to do that with sealed classes unfortunately, as far as I know at least

sdk/src/main/java/co/omise/android/models/Serializer.kt
Namiii: I'll see if I can make it more specific 🙆‍♂

sdk/src/main/java/co/omise/android/models/Source.kt
Namiii: While I agree that when I looked at this at the beginning, it's very messy and not optimal and having it as individual subclasses would certainly help with this. 
But having said that, I can also see how doing that would make the code more complex and less flexible and that could be the reason we have kept it that way in `omise-java`. Once I'm back next week, we can discuss on what's the best approach for improving this.

sdk/src/test/resources/data/objects/source_object.json
Namiii: The serialization test is broken anyway, I'll need to revamp it in the testing task.

Namiii:
@nuxzero changes made, please check. 

@pitiphong-p I've left some feedback on your comment and we'll address it once I'm back, please let me know if there are any other changes I need to apply for now.

sdk/src/test/resources/data/objects/source_object.json
pitiphong-p: You mean you need to do it in another PR?

sdk/src/main/java/co/omise/android/models/Source.kt
pitiphong-p: That's just my suggestion. I think we better stick to the convention of the language. Could you check how normally Kotlin devs do this similar things?

Namiii: Yeah I understand, I'm just saying I'd like us to take a little bit of time to figure out how to best improve this.

pitiphong-p: OK, let's discuss about this later

Namiii: 🙇

sdk/src/test/resources/data/objects/source_object.json
Namiii: Yeah there was already a task to complete testing for Requests. During this task I realized that the serialization test was actually broken so I add this task -> **T14297** as a subtask of the testing task.

PR #69, opened 2019-08-08
Theming credit card form

nuxzero:
### Objective

From previous PRs, we provided basically function for creating the Token via the `CreditCardActivity`. So, this PR continue theming the `CreditCardActivity` to provide the theme and able to custom the theme. 

### Description of changes

On this PR, I renamed the default theme from `OmiseSDKTheme` to better name `OmiseTheme`. Android provided style follow the design.

Custom theme, this PR declared new attributes to able set styles on the `CreditCardActivity`. After the user set styles to attributes on their theme, the layout should apply the set of styles via declared attributes.

```xml
<attr name="omiseButtonStyle" format="reference" />
<attr name="omiseEditTextStyle" format="reference" />
<attr name="omiseEditTextLabelStyle" format="reference" />
<attr name="omiseEditTextErrorStyle" format="reference" />
```

### QA

In this PR has the `SampleTheme` to show how to create a custom theme to use in the user's app. You can change the theme on the `AndroidManifest` to see how it looks on the sample app.

**Example theme**

```xml
 <style name="SampleTheme" parent="Theme.MaterialComponents">
        <item name="android:windowBackground">@color/primary</item>
        <item name="colorPrimary">@color/primary</item>
        <item name="colorPrimaryDark">@color/primary_dark</item>
        <item name="colorAccent">@color/accent</item>
        <item name="omiseEditTextStyle">@style/SampleEditText</item>
        <item name="omiseEditTextLabelStyle">@style/SampleEditTextLabel</item>
        <item name="omiseEditTextErrorStyle">@style/SampleEditTextError</item>
        <item name="omiseButtonStyle">@style/SampleButton</item>
</style>
```

OmiseTheme
SampleTheme

### DIFF
--- a/sdk/src/main/res/values/attrs.xml
+++ b/sdk/src/main/res/values/attrs.xml
@@ -1,8 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-
-    <attr name="omiseButtonStyle" format="reference" />
-    <attr name="omiseEditTextStyle" format="reference" />
-    <attr name="omiseEditTextLabelStyle" format="reference" />
-    <attr name="omiseEditTextErrorStyle" format="reference" />
+    <attr name="editTextLabelStyle" format="reference" />
+    <attr name="editTextErrorStyle" format="reference" />
 </resources>
--- a/sdk/src/main/res/values/attrs.xml
+++ b/sdk/src/main/res/values/attrs.xml
@@ -4,6 +4,7 @@
     <!-- EditText -->
     <attr name="omiseEditTextBackground" format="reference" />
     <attr name="omiseEditTextAppearance" format="reference" />
+    <attr name="omiseEditTextLabelAppearance" format="reference" />
     <attr name="omiseEditTextErrorAppearance" format="reference" />
 
     <!-- Button -->
--- a/sdk/src/main/res/values/attrs.xml
+++ b/sdk/src/main/res/values/attrs.xml
@@ -1,13 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
-    <!-- EditText -->
-    <attr name="omiseEditTextBackground" format="reference" />
-    <attr name="omiseEditTextAppearance" format="reference" />
-    <attr name="omiseEditTextLabelAppearance" format="reference" />
-    <attr name="omiseEditTextErrorAppearance" format="reference" />
-
-    <!-- Button -->
-    <attr name="omiseButtonTextAppearance" format="reference" />
-    <attr name="omiseButtonBackground" format="reference" />
+    <attr name="omiseButtonStyle" format="reference" />
+    <attr name="omiseEditTextStyle" format="reference" />
+    <attr name="omiseEditTextLabelStyle" format="reference" />
+    <attr name="omiseEditTextErrorStyle" format="reference" />
 </resources>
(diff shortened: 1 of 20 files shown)


sdk/src/main/res/values/colors.xml
```diff
@@ -1,4 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <color name="error">#ef3526</color>
+    <color  name="primary_accent">#1a56f0</color>
```
pitiphong-p: space

sdk/src/main/res/values/attrs.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <attr name="omiseButtonStyle" format="reference" />
+    <attr name="omiseEditTextStyle" format="reference" />
+    <attr name="omiseEditTextLabelStyle" format="reference" />
```
pitiphong-p: Do we have an alternative for this approach? Is it ok to use the existed style/attribute?

sdk/src/main/res/values/colors.xml
nuxzero: 😂

sdk/src/main/res/values/attrs.xml
nuxzero: 1. For setting the style via styles.xml seems to be only one way that think out for now. 
2. I think it betters existed style from the API cause the API it too general and I can't guarantee it'll not affect to another component.

pitiphong-p:
Can we provide a default theme value for the credit card form? I couldn't see it anywhere in the changes. I think we should have it since there's a very high chance that a merchant wouldn't set a theme with the omise* attributes

nuxzero:
I not quite sure to declare the **CreditCardActivity** with **OmiseTheme** to SDK's **AndroidManifest**. If we do that we have to declare all Activity that users may not use all of that. However if we OK to do that I'll make it. 🙂

pitiphong-p:
Do you have an example of a project that didn't set the theme?

sdk/src/main/res/values/colors.xml
Namiii: :)

nuxzero:
Here is a sample without setting theme. If user didn't set the theme it will choose Application's theme.

Screenshot_1565248146

pitiphong-p:
But the application theme wouldn't have a value for omise* attributes? Do I understand that point correctly?

pitiphong-p:
My point is that should we use the default attributes instead of introducing new ones? And if that's not the case then can we provide a default value for those attributes?

sdk/src/main/res/values/strings.xml
```diff
@@ -1,7 +1,7 @@
 <resources>
     <string name="not_available">(n/a)</string>
-    <string name="button_submit">Submit</string>
-    <string name="default_form_title">Add Credit Card</string>
+    <string name="button_submit">Pay</string>
```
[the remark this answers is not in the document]
nuxzero: :okay:

nuxzero:
I got you point. However I still need new attributes for prevent unexpected styles from system attribute.

What I can do for this is change `EditText` and `Button` to use system attribute and provide default styles via `OmiseTheme` but for `TextViews` may still need to declare new attributes.

pitiphong-p:
That sounds better to me. @Namiii what do you think?

Namiii:
@pitiphong-p @nuxzero 👌

nuxzero:
@pitiphong-p @Namiii @igroomgrim I had changed follow discussed above. Could help me review 2 last commits please? 🙏

PR #73, opened 2019-08-27
Circle ci integration

Namiii:
**TASK**=> T13623

**SUMMARY**
Added CircleCi configuration 

👾**Note** 👾
We are still not getting a pass from CircleCi but that is because there are failing tests and not because there is no configuration.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,27 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/omise/omise-android
+    docker:
+      - image: circleci/android:api-28
+    environment:
+      JVM-OPTS: -Xmx3200m
+    steps:
+      - checkout
+      - restore_cache:
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,27 @@
...
+            - ~/.gradle
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
+      - run:
+          name: Run Tests
+          command: ./gradlew lint test
+      - store_artifacts:
+          path: app/build/reports
+          destination: reports
+      - store_test_results:
+          path: app/build/test-results
```
nuxzero: Could you use SDK path instead sample app path?

Namiii: As in `~/` ?

nuxzero: I think that represent to root path. exactly I asking for use `sdk` instead of `app`.

nuxzero: Or maybe it can add multiple path.

Namiii: Ahhhh now I understand what you mean 🤦‍♂ I'll change it

PR #74, opened 2019-08-28
Integrate Parcelize in the project

Namiii:
**TASK**=> T13872

**SUMMARY**
Started using `@Parcelize` annotation for model classes in the SDK in order to reduce the `Parcelable` implementation boilerplate code.

**NOTE**
As part of this task, we had originally planned to change the way model properties were set up but that has been moved into its own task => T16075


sdk/src/main/java/co/omise/android/models/FlowType.kt
```diff
@@ -1,26 +1,42 @@
...
  * @see [Sources API](https://www.omise.co/sources-api)
  */
 sealed class FlowType(
-        val name: String
+        val name: String?
 ) {
 
     object Redirect : FlowType("redirect")
     object Offline : FlowType("offline")
+    object Default : FlowType(null)
```
pitiphong-p: `Default` looks misleading to me. I think we should choose a difference name for this?

sdk/build.gradle
```diff
@@ -29,6 +29,10 @@ android {
         disable 'InvalidPackage' // okio
     }
 
+    androidExtensions {
+        experimental = true
```
pitiphong-p: Experimental? When will the feature we use is out of the experimental phase?

Namiii: It's been flagged as `experimental` for a while now, about a year? @nuxzero correct me if I'm wrong?

sdk/src/main/java/co/omise/android/models/FlowType.kt
Namiii: But I'm literally using it as Default for Jackson, what could we replace it with? `Unknown`? `Invalid`?

pitiphong-p: Could it be `Unknown(value)` instead of null?

Namiii: No unfortunately, not with sealed classes. In any event, when Unknown is used in our code, there is no value to pass in anyway. I'll change the name to `Unknown` 👌

Namiii:
@pitiphong-p Change done 🙆‍♂

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
```diff
@@ -0,0 +1,96 @@
+package co.omise.android
+
+import android.os.Bundle
+import android.os.Parcel
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import co.omise.android.models.*
```
nuxzero: Could you use single import instead of wildcard import?

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
```diff
@@ -0,0 +1,96 @@
...
+import org.joda.time.DateTime
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class ParcelableTest {
+
+    @Test
+    fun testModelParceling() {
```
nuxzero: Honestly this name bring me back to JUnit 3. Could you rename them?

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        }
+        val capability = Capability(
+                mutableListOf("a", "b", "c", "d"),
+                paymentMethodList,
+                false)
+
+        testObjectParceling(capability)
+    }
+
+    private fun <T : Model> testObjectParceling(model: T) {
```
nuxzero: Is this function to assert the Parcel object right? What about this name `assertParcelObject`?

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
```diff
@@ -0,0 +1,96 @@
...
+import android.os.Parcel
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import co.omise.android.models.*
+import org.joda.time.DateTime
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class ParcelableTest {
```
nuxzero: I believe this test class can run in unit test. Could you check it?

sdk/build.gradle
nuxzero: Sorry, I don't what actually when it will release to stable version but I think they will not release in soon. 

Q: Should we use it?
A: I remember there is an important note about using experimental mode but it gone from the website. So, case of building the SDK I think it quite risky to the SDK because it might raise to bugs in a head since we need to support a ton of variants (e.g. Java, Kotlin) that I'm quite concern. But we have test enough to cover all areas that we use it that might acceptable. 😂

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
Namiii: No it cannot as `Parcel` is part of the Android sdk

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
Namiii: 👍

Namiii: After importing all one by one, the IDE is auto correcting it to wild card, so I guess it's best to leave it that way.

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
Namiii: Not sure what you mean by JUnit 3 naming, what would we change it to?

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
Namiii: Yeah that makes sense

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
nuxzero: Are you sure? Since we setup Robolectric that can include all Android resources into Unit test. I think every thing that you can run with `AndroidJUnit4` it should runnable in unit test. 🤔

sdk/build.gradle
Namiii: I feel it's fine to use, but again we can keep testing it to see if there are any issues. If there are, it's not that hard to undo the changes.

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
Namiii: We have? When did that happen? I have zero experience with Roboelectric, so please let me know how this can be changed into a unit test 👍

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
nuxzero: I think you change it from the setting.

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
nuxzero: Could you check from other test classes?

nuxzero: Remember we discussed about this before. 😂

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
Namiii: We can, but if we are import 8/14 files in a package, it does kind of make sense to have this rule to use wildcard import

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
Namiii: We did, you mean like `testModelParceling_Success`? I'll change them but I don't see the point adding that at the end just for the sake of adding it, since there are no different states for each test, they are all success states. Will make the changes

nuxzero: Since JUnit 4 release it use `@Test` instead of put `test` prefix (it uses in JUnit 3) in test cases. So, it no point to put `test` as prefix.

Namiii: I see, ok I'll make those changes too 👌

Namiii: Done

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
Namiii: Done 👌

Namiii:
@nuxzero all changes made

sdk/src/androidTest/java/co/omise/android/ParcelableTest.kt
nuxzero: So, how many import that we start to use wildcard?

PR #77, opened 2019-08-30
Fix deserialize source response

nuxzero:
### Objective

To fix the source response failed to deserialize `type` and `flow` cause the response not match with the instance creator function.

### Description of code changes

Corrected checking response field value and return `Unknown` with raw value if not match any `SourceType` or `FlowType`.

### DIFF
--- a/sdk/src/main/java/co/omise/android/models/FlowType.kt
+++ b/sdk/src/main/java/co/omise/android/models/FlowType.kt
@@ -12,29 +12,29 @@ import kotlinx.android.parcel.Parceler
  * @see [Sources API](https://www.omise.co/sources-api)
  */
 sealed class FlowType(
-        @JsonValue val name: String?
+        @JsonValue open val name: String?
 ) {
 
     object Redirect : FlowType("redirect")
     object Offline : FlowType("offline")
-    object Unknown : FlowType(null)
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/co/omise/android/models/FlowType.kt
```diff
@@ -12,29 +12,29 @@ import kotlinx.android.parcel.Parceler
...
  */
 sealed class FlowType(
-        @JsonValue val name: String?
+        @JsonValue open val name: String?
 ) {
 
     object Redirect : FlowType("redirect")
     object Offline : FlowType("offline")
-    object Unknown : FlowType(null)
+    data class Unknown(override val name: String?) : FlowType(name)
```
pitiphong-p: THIS IS WHAT I WANT!!! Thank you

Namiii: 😂😂😂

nuxzero: I need it too.

Namiii: Yeah I'm glad you found a way to make it work 👏

nuxzero:
Yeah, it's mix together 🙂

PR #79, opened 2019-09-05
Fix base class serialization

Namiii:
**TASK**=>T16279

**SUMMARY**
The properties in the base class, `Model`, were not being deserialized by Jackson, so they were converted into `Interface` properties and implement that interface on all the Model classes.

### DIFF
--- a/sdk/src/main/java/co/omise/android/models/BaseModel.kt
+++ b/sdk/src/main/java/co/omise/android/models/BaseModel.kt
@@ -1,0 +1,8 @@
+package co.omise.android.models
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver
+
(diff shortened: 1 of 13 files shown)


sdk/src/main/java/co/omise/android/models/BaseModel.kt
```diff
@@ -0,0 +1,8 @@
+package co.omise.android.models
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.CUSTOM, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "object", visible = true)
+@JsonTypeIdResolver(ModelTypeResolver::class)
```
nuxzero: Is it possible to put this annotations to `Model`?

Namiii: From memory, I don't think so. But I'll double check and let you know

Namiii: @nuxzero it actually still works, so made the change 👌

PR #87, opened 2019-09-13
Create payment from installment

nuxzero:
### Objective

Create the installment chooser page to able users to create a payment from the installment method.

### Description of changes

In this PR create new 2 pages for the flow. The first page is `InstallmentChooserFragment` for choosing a bank issuer. 

And the second page is `InstallmentTermChooserFragment` for choosing an installment term. After the user chose an installment term, it will send a request to create a source object.

### QA

To test creating a source from the installment method, you can open the sample app and choose the installment method. Then choose a bank issuer and installment term it should return source object.

Screenshot_1568361166

Screenshot_1568361171

### DIFF
--- a/sdk/src/main/java/co/omise/android/models/SourceType.kt
+++ b/sdk/src/main/java/co/omise/android/models/SourceType.kt
@@ -36,6 +36,18 @@ sealed class SourceType(
         object Ktc : Installment("installment_ktc")
         object KBank : Installment("installment_kbank")
         data class Unknown(@JsonValue override val name: String?) : Installment(name)
+
+        companion object {
+            fun availableTerms(installment: Installment): List<Int> =
+                    when (installment) {
+                        Bay -> listOf(3, 4, 6, 9, 10)
+                        FirstChoice -> listOf(3, 4, 6, 9, 10, 12, 18, 24, 36)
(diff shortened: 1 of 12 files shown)


sdk/src/main/java/co/omise/android/models/SourceType.kt
```diff
@@ -36,6 +36,18 @@ sealed class SourceType(
         object Ktc : Installment("installment_ktc")
         object KBank : Installment("installment_kbank")
         data class Unknown(@JsonValue override val name: String?) : Installment(name)
+
+        companion object {
+            fun availableTerms(installment: Installment): List<Int> =
+                    when (installment) {
+                        Bay -> listOf(3, 4, 6, 9, 10)
```
Namiii: Not sure if we have any other options but this is gonna make our code very inflexible

nuxzero: That right, this for the case that the user want to create their own config rather than using installment terms from Capability API. So, this will help to not make a mistake from set the unavailable installment terms. 🙂

Namiii: 👌 👌 👌

PR #88, opened 2019-09-16
Create payment from EContext

nuxzero:
### Objective

To able users create a payment from Convenience Store / Pay-easy / Netbanking in the payment chooser page.

### Description of changes

When the user pressed Convenience Store / Pay-easy / Netbanking in the payment chooser page, it will display a form that required full-name, email, phone number. When the user pressed the next button, it will send a request to create a source with information in the form.

### QA

When the user entered valid information, it will send a request and return a source object.

If there is invalid information, the submit button will disable to press til the user enter valid data.

create-source-from-econtext

### DIFF
--- a/sdk/src/main/java/co/omise/android/ui/EContextFormFragment.kt
+++ b/sdk/src/main/java/co/omise/android/ui/EContextFormFragment.kt
@@ -11,6 +11,7 @@ import android.widget.TextView
 import co.omise.android.R
 import co.omise.android.extensions.setOnAfterTextChangeListener
 import co.omise.android.extensions.setOnClickListener
+import co.omise.android.models.SupportedEContext
 import co.omise.android.models.Source
 import co.omise.android.models.SourceType
 import kotlinx.android.synthetic.main.fragment_econtext_form.button_submit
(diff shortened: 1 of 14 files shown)


sdk/src/main/java/co/omise/android/ui/EContextFormFragment.kt
```diff
@@ -0,0 +1,125 @@
...
+        val email = emailEdit.text?.toString()?.trim().orEmpty()
+        val phoneNumber = phoneNumberEdit.text?.toString()?.trim().orEmpty()
+
+        val request = Source.CreateSourceRequestBuilder(requester.amount, requester.currency, SourceType.Econtext)
+                .name(fullName)
+                .email(email)
+                .phoneNumber(phoneNumber)
+                .build()
+
+        view?.let { setAllViewsEnabled(it, false) }
```
Namiii: Can the `let` be replace by a direct call on `view?`?

nuxzero: Sorry, `setAllViewsEnabled()` couldn't pass the nullable `view`. 😄

PR #95, opened 2019-09-25
Payment setting page

nuxzero:
### Objective

To able users to adjust payment options via the payment setting page on the sample app.

### Description of changes

In this PR created the payment setting page with setting options. That allowed users to choose payment setting e.g., use payment methods from the capability API or use custom payment methods,  select payment methods. When the user press **Choose Payment Method** button, it will choose the capability data by the user's setting and pass the data to the payment chooser page.

### QA

When the user had changed payment settings, the payment chooser page will change payment methods by following the user's settings.
Screenshot_1569403595

Screenshot_1569403598


app/build.gradle
```diff
@@ -40,9 +40,9 @@ dependencies {
     implementation "androidx.multidex:multidex:$multidex_version"
     implementation "androidx.appcompat:appcompat:$appcompat_version"
     implementation "androidx.constraintlayout:constraintlayout:$constraintlayout_version"
+    implementation "androidx.preference:preference:$preference_version"
     implementation "com.google.android.material:material:$material_version"
     implementation "com.google.guava:guava:$guava_version"
-    implementation "com.squareup.picasso:picasso:$picasso_version"
```
Namiii: Why did we have Picasso? :)

nuxzero: haha, I forgot to delete it from the previous version of the sample app. 😅

PR #98, opened 2019-10-01
Translate payment creator

nuxzero:
In this PR translated the SDK to support Thai and Japan.

### DIFF
--- a/sdk/src/main/res/values/strings.xml
+++ b/sdk/src/main/res/values/strings.xml
@@ -4,7 +4,6 @@
     <string name="button_next">Next</string>
     <string name="default_form_title">Credit Card</string>
     <string name="description_brand_logo">Card brand logo</string>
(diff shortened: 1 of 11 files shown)


sdk/src/main/res/values/strings.xml
```diff
@@ -61,6 +59,29 @@
     <string name="payment_chooser_title">Payment Methods</string>
     <string name="internet_banking_chooser_title">Internet Banking</string>
     <string name="installments_title">Installments</string>
-    <string name="econtext_title">EContext</string>
+    <string name="econtext_title">econtext</string>
```
igroomgrim: That use all lowercase?

nuxzero: Yeah, it seems to be like that according to our copy writer.

PR #110, opened 2019-10-28
Api request log

nuxzero:
In this PR put API request logs in the debug mode to help users checking API request.


sdk/src/main/java/co/omise/android/SDKLog.kt
```diff
@@ -0,0 +1,25 @@
...
+        }
+    }
+
+    fun e(message: String, e: Throwable) {
+        if (BuildConfig.DEBUG) {
+            Log.e(TAG, message, e)
+        }
+    }
+
+    fun wtf(message: String, e: Throwable) {
```
igroomgrim: What did you mean here? What the 🦊 😄

nuxzero: **What a Terrible Failure** 😂 
https://developer.android.com/reference/android/util/Log#wtf(java.lang.String,%20java.lang.Throwable)

Namiii: :D

PR #128, opened 2020-04-14
T20531 - Fix creating source

nuxzero:
**1. Objective**

To fix creating source received a response that contains unrelated the parameters.

**2. Description of change**

In this PR rid of default value from the create source parameters to prevent create a source with default field values.

**3. Quality assurance**

When create a new source, the response shouldn't contains fields that unrelated to the parameters.

**4. Operations impact**

N/A

**5. Business impact**

N/A

**6. Priority of change**

Normal

### DIFF
--- a/sdk/src/main/java/co/omise/android/models/Source.kt
+++ b/sdk/src/main/java/co/omise/android/models/Source.kt
@@ -79,7 +79,7 @@ data class Source(
         @JsonProperty("phone_number")
         private var phoneNumber: String? = null
         @JsonProperty("installment_term")
-        private var installmentTerm: Int = 0
+        private var installmentTerm: Int? = null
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/co/omise/android/models/Source.kt
```diff
@@ -79,7 +79,7 @@ data class Source(
         @JsonProperty("phone_number")
         private var phoneNumber: String? = null
         @JsonProperty("installment_term")
-        private var installmentTerm: Int = 0
+        private var installmentTerm: Int? = null
```
igroomgrim, line 82: Cause it can be null from our API right?

nuxzero: Yup.

Issue #164, opened 2020-11-20
As an Omise developer, I want to add a new source type in minimum places


nuxzero:
# As an Omise developer, I want to add a new source type in minimum places

## Story

As an Omise developer, I want to add the new payment method in the payment form page of the Omise Android SDK. But it’s slightly complicated because there are many places required for me to change in the codebase.

## Proposal

To accomplish this story, the Omise Android SDK shall contain all resources of payment methods in a single source. And the Omise Android SDK shall consume the resources for displaying in the payment creator page. 

**Declaration**

```kotlin
sealed class PaymentMethodResource(...) {
  object CreditCard(
    iconRes = R.drawable.payment_card,
    titleRes = R.string.payment_method_credit_card_title,
    indicatorIconRes = R.drawable.ic_next
  ): PaymentMethodResource

  object InstallmentBbl(
    iconRes = R.drawable.payment_bbl,
    titleRes = R.string.payment_method_installment_bbl_title,
    indicatorIconRes = R.drawable.ic_next,
    sourceType = InstallmentChooserItem.Bbl
  ): PaymentMethodResource

  object InternetBankingScb(
    iconRes = R.drawable.payment_scb,
    titleRes = R.string.payment_method_internet_banking_scb_title,
    indicatorIconRes = R.drawable.ic_redirect,
    sourceType = SourceType.InternetBanking.Scb
  ): PaymentMethodResource
}
```

**Usage**
```kotlin
class OmiseItemViewHolder(val view: View, val listener: OmiseListItemClickListener?) : RecyclerView.ViewHolder(view) {
  fun bind(item: PaymentMethodResource) {
    ... 
    optionImage.setImageResource(item.iconRes)
    optionImage.visibility = View.VISIBLE
    nameText.text = item.titleRes
    typeImage.setImageResource(item.indicatorIconRes)
    view.setOnClickListener { listener?.onClick(item) }
  }
}
```

PR #165, opened 2020-11-24
Revamp payment method resources

nuxzero:
## Purpose

This purpose to gather payment method resources that using in the **PaymentCreatorActivity**'s fragments to a single place.

**Issue:** #164 

## Description

This PR introduced the **PaymentMethodResources** that contains all UI resources about the payment methods for using in the **PaymentCreatorActivity**'s fragments. Also, refactored the **PaymentCreatorActivity**'s fragments to consume UI resources from the **PaymentMethodResources**.

## Quality assurance

The payment creator page shall perform as the same behave.

## Operations impact

N/A

## Pre- and post-deployment steps

N/A

## Business impact (release notes)

N/A

## Add labels and assign reviewers

N/A

## Back-out Procedure

N/A


igroomgrim (approved):

tookaa2 (approved):

sdk/src/main/java/co/omise/android/models/Capability.kt
```diff
@@ -66,3 +66,18 @@ data class Capability(
         }
     }
 }
+
```
danfowler, line 69: General point about `/capability`.  The latest version of our API looks like this (e.g. has a `payment_methods` attribute), but previous versions are incompatible (e.g. they have a `payment_backends` attribute).  The data structure is also different.

danfowler: Oh, and I guess we are setting the API version to be 2019-05-29 so it's OK 👍

danfowler (approved):

nikolack (approved):

sdk/src/main/java/co/omise/android/models/Capability.kt
nuxzero: Correct. 👍

PR #191, opened 2021-02-22
Update test workflow

nikolack:
## 1. Objective
Enable automated UI testing for Omise-Android and enable test reporting

## 2. Description of change
-UI testing is already enabled through Robotium framework
-Added test reports. After every test, report will be generated in the `Artifacts` section where detailed report of failing tests can be found.
-Changed the way the tests are triggered on Github, so that the Github Actions doesn't run same actions twice on every push.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,5 +45,5 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v2
         with:
-          name: my-artifact
+          name: test-report
           path: sdk/build/reports
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@
 
 name: Android CI
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


nuxzero (approved):
Note: Some UI tests are not working with Robolectric. I think in the future we might need to fix those tests to support testing with Robolectric. 🙂

.github/workflows/ci.yml
```diff
@@ -41,3 +41,9 @@ jobs:
           PGP_SIGNING_PASSWORD: ${{ secrets.PGP_SIGNING_PASSWORD }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+      - name: Upload test report
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: my-artifact
```
tobias90, line 48: Maybe you have a more meaningful name for the artifact than `my-artifact`?

nikolack: Changed

tobias90 (approved):

danfowler (approved):

PR #230, opened 2022-10-06
Release v4.3.0

new4762:
## v4.3.0
* **ADDED:** Added support ShopeePay JumpApp.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         sample_app_version = '1.0'
         sample_app_code_version = 1
 
-        omise_sdk_version = '4.2.1'
+        omise_sdk_version = '4.3.0'
         omise_sdk_code_version = 30
 
         compile_sdk_version = 30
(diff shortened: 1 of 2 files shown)


tookaa2 (approved):

hpariwat (approved):

build.gradle
```diff
@@ -5,8 +5,8 @@ buildscript {
         sample_app_version = '1.0'
         sample_app_code_version = 1
 
-        omise_sdk_version = '4.2.0'
-        omise_sdk_code_version = 29
+        omise_sdk_version = '4.2.1'
```
kan98, line 8: this looks more like a feature than a patch. So if we follow https://semver.org/ should it be `4.3.0` instead?

new4762: Good point here updated https://github.com/omise/omise-android/pull/230/commits/9050e751262f2699399f9dca520b8b1d58c9029a 🙏

kan98 (approved):

PR #239, opened 2022-12-22
Add Support for Atome

jutinapas:
* Add `atome` as a new payment method
* Add support on new fields in Source API for BNPL requests with a related user input form in the Example App

Screenshot_20221223_100940 1 (1) Screenshot_20221222_173554 1 (2) Screenshot_20221223_101216 1 (1)

image

PR #240, opened 2023-01-04
Release 4.5.0

jutinapas:
## v4.5.0

* **ADDED:** Added support for Atome payment method

Related PR: https://github.com/omise/omise-android/pull/239

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         sample_app_version = '1.0'
         sample_app_code_version = 1
 
-        omise_sdk_version = '4.4.3'
+        omise_sdk_version = '4.5.0'
         omise_sdk_code_version = 35
 
         compile_sdk_version = 30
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -5,8 +5,8 @@ buildscript {
         sample_app_version = '1.0'
         sample_app_code_version = 1
 
-        omise_sdk_version = '4.4.2'
-        omise_sdk_code_version = 34
+        omise_sdk_version = '4.4.3'
```
kan98, line 8: since it's a feature, can we do `4.5.0` instead? to follow semantic versioning https://semver.org/

kan98: otherwise our merchants setup may do an auto update as they will think it's a patch/bug fix.

jutinapas: Thanks! Fixed here instead -> https://github.com/omise/omise-android/pull/241

PR #245, opened 2023-01-12
[EM-2969] Remove KTB and SCB internet bankings

Narum11:
To remove KTB and SCB internet bankings, regarding the SCB and KTB internet banking decommission.

Task: https://opn-ooo.atlassian.net/browse/EM-2786

Results: 
SCB and KTB internet banking do not appear in payment methods selection both user setting page and payment page.

Screen Shot 2023-01-12 at 2 51 46 PM   Screen Shot 2023-01-12 at 2 52 07 PM

### DIFF
--- a/app/src/main/java/co/omise/android/example/PaymentSetting.kt
+++ b/app/src/main/java/co/omise/android/example/PaymentSetting.kt
@@ -15,8 +15,6 @@ object PaymentSetting {
                     R.string.payment_preference_zero_interest_installments_key,
                     R.string.payment_preference_credit_card_key,
                     R.string.payment_preference_internet_banking_bay_key,
-                    R.string.payment_preference_internet_banking_ktb_key,
-                    R.string.payment_preference_internet_banking_scb_key,
                     R.string.payment_preference_internet_banking_bbl_key,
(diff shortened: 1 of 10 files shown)


hpariwat (approved):

jutinapas (approved):

FhanOmise (approved):

kan98 (approved):

tookaa2 (approved):

nuxzero:
Removing payment methods might break the merchant apps once they adopted this version. Perhaps, we need to mark them as the payment methods are deprecated and doc how's they can migrate to the new payment methods. 
I also have a question about backend, are these payment methods going to deprecate from the backend side as well?

Narum11:
> Removing payment methods might break the merchant apps once they adopted this version. Perhaps, we need to mark them as the payment methods are deprecated and doc how's they can migrate to the new payment methods. I also have a question about backend, are these payment methods going to deprecate from the backend side as well?

Of course, we also deprecate these payment methods in our backend capability api, and we will remove both methods out from supported banks in omise document as well ka.

nuxzero (approved):

PR #247, opened 2023-01-20
[CS-1848] Add browser flow option

nikolack:
Enable browser flow for AuthorizingPaymentActivity. This will open authorize_url in browser, instead of in-app.

CS Ticket: https://opn-ooo.atlassian.net/browse/CS-1804
https://opn-ooo.atlassian.net/browse/CS-1848


tookaa2 (approved):

aashishgurung:
Can you please include the Jira ticket?

nikolack:
@aashishgurung  I don't have the Jira ticket. @tookaa2  can you please provide?

tookaa2:
@aashishgurung @nikolack we have this ticket which also links to Engagement's ticket
https://opn-ooo.atlassian.net/browse/CS-1804

tookaa2:
here is the ticket @aashishgurung @nikolack
https://opn-ooo.atlassian.net/browse/CS-1846

new4762 (approved):

aashishgurung (approved):

nuxzero:
Not sure, do we really want to support open external browser because they can implement by their own without using the SDK. However, Opening the external browser might required the app to support deeplink in order to redirect back to the app.

nikolack:
@nuxzero not sure what could be the alternative solution. From the docs it seems like Bangkok bank app needs to whitelist app in order to navigate there using `startActivityForResult`

kan98 (approved):

nikolack:
@nuxzero I think the solution you proposed today (to check the authorize URL and if it belongs to bangkok bank, open browser). This way there is no need for merchant to make any change to their app. It will affect all merchants using our app though.

PR #261, opened 2023-06-07
[MIT-1378] Upgrade 3DS-SDK version 1.0.0-alpha12

nuxzero:
This PR upgraded `threeds-android` dependency to vresion 1.0.0-alpha12. 

**Minimum requirements**
- Kotlin version 1.7.10 or above.
- Android gradle plugin version 7.4.2 or above.

### DIFF
--- a/sdk/src/main/java/co/omise/android/ui/PaymentChooserFragment.kt
+++ b/sdk/src/main/java/co/omise/android/ui/PaymentChooserFragment.kt
@@ -52,13 +52,13 @@ internal class PaymentChooserFragment : OmiseListFragment<PaymentMethodResource>
             PaymentMethodResource.MaybankQR,
             PaymentMethodResource.GrabPay,
             PaymentMethodResource.PayPay,
-            PaymentMethodResource.PointsCiti -> item.sourceType?.let(::sendRequest)
+            PaymentMethodResource.PointsCiti,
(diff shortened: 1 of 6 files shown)


danfowler (approved):

aashishgurung (approved):

sdk/src/main/java/co/omise/android/ui/PaymentChooserFragment.kt
```diff
@@ -52,7 +52,9 @@ internal class PaymentChooserFragment : OmiseListFragment<PaymentMethodResource>
             PaymentMethodResource.MaybankQR,
             PaymentMethodResource.GrabPay,
             PaymentMethodResource.PayPay,
-            PaymentMethodResource.PointsCiti -> item.sourceType?.let(::sendRequest)
+            PaymentMethodResource.PointsCiti,
+            PaymentMethodResource.GrabPay_RMS,
+            PaymentMethodResource.TouchNGo_Alipay -> item.sourceType?.let(::sendRequest)
```
danfowler, line 57: @nuxzero just to confirm, `-> item.sourceType?.let(::sendRequest)` should follow the last item which used to be `PaymentMethodResource.PointsCiti` but now is `PaymentMethodResource.TouchNGo_Alipay`, correct?

In that case, should we either define this as an `else` clause or somehow group these payment methods together?

PR #265, opened 2023-06-15
[MIT-1081] Support billing address for AVS countries

nuxzero:
This PR enhances the credit card form to support billing address for US, UK, and Canadian cardholders.
https://www.omise.co/How-to-improve-my-authorization-rate-for-US-UK-and-Canadian-cardholders


sdk/src/main/java/co/omise/android/ui/CreditCardActivity.kt
```diff
@@ -16,61 +18,145 @@ import co.omise.android.api.RequestListener
...
+                invalidateBillingAddressForm()
+            }
+        }
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
         setContentView(R.layout.activity_credit_card)
 
+        require(intent.hasExtra(EXTRA_PKEY)) { "Could not found ${::EXTRA_PKEY.name}." }
```
danfowler, line 130:
```suggestion
        require(intent.hasExtra(EXTRA_PKEY)) { "Could not find ${::EXTRA_PKEY.name}." }
```

AnasNaouchi (commented):
should we have a predefined list of zip codes that is associated to the list of countries or that is not the intended behavior ? @danfowler @nuxzero

nuxzero:
It's not scope of this PR but it's nice to have. Probably we can discuss and make an enhancement in the future.

nuxzero:
> It's not scope of this PR but it's nice to have. Probably we can discuss and make an enhancement in the future.



> should we have a predefined list of zip codes that is associated to the list of countries or that is not the intended behavior ? @danfowler @nuxzero

It's not scope of this PR but it's nice to have. Probably we can discuss and make an enhancement in the future.

danfowler (approved):

AnasNaouchi (approved):

app/src/java/java/co/omise/android/example/CheckoutActivity.java
```diff
@@ -240,13 +240,16 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
             if (data.hasExtra(OmiseActivity.EXTRA_SOURCE_OBJECT)) {
                 Source source = data.getParcelableExtra(OmiseActivity.EXTRA_SOURCE_OBJECT);
                 snackbar.setText(source.getId()).show();
+                Log.d(TAG, "source: " + source.getId());
```
danfowler, line 243: We're logging on purpose here for the sake of the example app?

danfowler (approved):

app/src/java/java/co/omise/android/example/CheckoutActivity.java
nuxzero: Yes, just in case we want to use the token to create a charge.

PR #269, opened 2023-09-07
[MIT-1510] Coverage report

nuxzero:
- Run instrumentation tests to CI flow
- Create coverage report to SonarCloud (43.9% coverage)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,6 +1,6 @@
 name: Code Coverage
 
-on: [ pull_request ]
+on: push
 
 jobs:
   coverage-report:
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # required for SonarCloud
 
       - name: Set up JDK 17
         uses: actions/setup-java@v3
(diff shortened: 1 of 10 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,57 @@
+name: Code Coverage
+
+on: [ pull_request ]
```
danfowler, line 3: @nuxzero @ajzkk @aashishgurung I noticed that some of our other public libraries trigger this action on push:

https://github.com/omise/omise-woocommerce/blob/develop/.github/workflows/code-coverage.yml#L3-L4

What should be the trigger?

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  coverage-report:
+    name: Coverage Report
+    runs-on: macos-latest
```
danfowler, line 8: @nuxzero can you leave a note on why we currently need `macos-latest` instead of `ubuntu-latest`?

.github/workflows/coverage.yml
nuxzero: Good point, I used `pull_request` to keep it same as the test workflow.

aashishgurung: I would say `push` because we need to run this action when something is pushed.

nuxzero: @danfowler @aashishgurung I updated the trigger to `push`.

danfowler (approved):

aashishgurung (approved):

PR #282, opened 2023-10-26
Enable webview for legacy payments

AnasNaouchi:
Enable webview for legacy payments that have `/pay` as ending route in auth url
- Additional result code is introduced to allow integrators to determine if the authentication terminated from webview: 
```
const val WEBVIEW_CLOSED_RESULT_CODE = 5
```

### DIFF
--- a/sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
+++ b/sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
@@ -56,21 +56,8 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_authorizing_payment)
-
-        supportActionBar?.title = uiCustomization.uiCustomization.toolbarCustomization?.headerText
-            ?: getString(R.string.title_authorizing_payment)
-
-        val authUrl = verifier.authorizedURLString
-        // check for legacy payments that require web view
-        if (authUrl.endsWith("/pay")) {
-            setupWebView()
-        } else {
-            // Check if the URL needs to be opened externally
-            if (verifier.verifyExternalURL(verifier.authorizedURL)) {
-                openDeepLink(verifier.authorizedURL)
-            }
-            observeData()
-        }
+        setupActionBarTitle()
+        handlePaymentAuthorization()
     }
 
     @TestOnly
@@ -185,6 +172,25 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
         }
     }
 
+    private fun setupActionBarTitle() {
+        supportActionBar?.title = uiCustomization.uiCustomization.toolbarCustomization?.headerText
+            ?: getString(R.string.title_authorizing_payment)
+    }
+
+    private fun handlePaymentAuthorization() {
+        val authUrlString = verifier.authorizedURLString
+        val authUrl=verifier.authorizedURL
+        // check for legacy payments that require web view
+        if (authUrlString.endsWith("/pay")) {
+            setupWebView()
+        } else {
+            // Check if the URL needs to be opened externally
+            if (verifier.verifyExternalURL(authUrl)) {
+                openDeepLink(authUrl)
+            }
+            observeData()
+        }
+    }
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
         super.onActivityResult(requestCode, resultCode, data)
         if (requestCode == REQUEST_EXTERNAL_CODE) {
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -59,11 +60,17 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
         supportActionBar?.title = uiCustomization.uiCustomization.toolbarCustomization?.headerText
             ?: getString(R.string.title_authorizing_payment)
 
-        if (verifier.verifyExternalURL(verifier.authorizedURL)) {
-            openDeepLink(verifier.authorizedURL)
+        val authUrl = verifier.authorizedURLString
+        // check for legacy payments that require web view
+        if (authUrl.endsWith("/pay")) {
```
ajzkk, line 65: Is this the only one possibility to verify legacy url?

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -59,11 +60,17 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
...
+            setupWebView()
+        } else {
+            // Check if the URL needs to be opened externally
+            if (verifier.verifyExternalURL(verifier.authorizedURL)) {
+                openDeepLink(verifier.authorizedURL)
+            }
+            observeData()
```
aashishgurung, line 72: These lines are for authorization? How about moving these to a separate function like below. We might also combine `if..else { if }` as `if...else if`.

```
override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_authorizing_payment)

        supportActionBar?.title = uiCustomization.uiCustomization.toolbarCustomization?.headerText
            ?: getString(R.string.title_authorizing_payment)

        handlePaymentAuthorization()
}

private fun handlePaymentAuthorization() {
    val authUrl = verifier.authorizedURLString
 
    if (isLegacyPayment(authUrl)) {
        setupWebView()
    } else if (verifier.verifyExternalURL(verifier.authorizedURL)) {
         openDeepLink(verifier.authorizedURL)
    }

    observeData()
}
```

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -59,11 +60,17 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
         supportActionBar?.title = uiCustomization.uiCustomization.toolbarCustomization?.headerText
             ?: getString(R.string.title_authorizing_payment)
```
aashishgurung, line 61: These can also be moved to a separate function and call the function from `onCreate`

```
private fun setupActionBarTitle() {
    val headerText = uiCustomization.uiCustomization.toolbarCustomization?.headerText
    supportActionBar?.title = headerText ?: getString(R.string.title_authorizing_payment)
}
```

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
AnasNaouchi: Yes from what @nuxzero described in the ticket.

ajzkk: I saw 2 ways from the ticket, which of the following options is potentially more precise?
- The string end with `/authorize` for payment v2
- The string end with `/pay` for legacy payments

AnasNaouchi: @ajzkk  Some legacy payments support payment v2. Payment v2 accepts receiving an authorization request and will not reply with `404`. V2 end with `/authorize` which we do not want to change. On the other hand some other legacy payments do not support payment v2. These urls end with `/pay` and these are the urls that we want to specifically handle

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -185,6 +172,25 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
...
+        // check for legacy payments that require web view
+        if (authUrlString.endsWith("/pay")) {
+            setupWebView()
+        } else {
+            // Check if the URL needs to be opened externally
+            if (verifier.verifyExternalURL(authUrl)) {
+                openDeepLink(authUrl)
+            }
+            observeData()
+        }
```
aashishgurung, line 192:
```suggestion
        } else if (verifier.verifyExternalURL(authUrl)) {
            openDeepLink(authUrl)
        }

        observeData()
```

AnasNaouchi: @aashishgurung  `observeData()` should not be launched in case of legacy payment

aashishgurung: How does our current logic handle that?

- if URL ends with `/pay`, setup web view
- if not, it will go to `else` block and opens deep link if authURL is verified as externalUrl.
- if not it will trigger observeData.

So, `observeData` will still be triggered in our current implementation.

AnasNaouchi: yes it should be triggered on all cases except legacy because previously `observe` was called on all cases so now it should be called on all except legacy

aashishgurung: Got it. Thanks for explaining

aashishgurung (approved):

PR #286, opened 2023-11-09
Fallback to web flow in case Netcetera 3DS SDK fails

AnasNaouchi:
In case the Netcetera SDk fails to start/initialize the webview will be used to open the authUrl directly.

### DIFF
--- a/sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
+++ b/sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
@@ -265,9 +265,9 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
         val resultIntent = Intent().apply {
             putExtra(EXTRA_AUTHORIZING_PAYMENT_RESULT, Failure(throwable))
         }
-        if(arrayOf("Challenge protocol error","Challenge runtime error","3DS2 initialization failed").contains(throwable.message)){
+        if (arrayOf("Challenge protocol error","Challenge runtime error","3DS2 initialization failed").contains(throwable.message)) {
             setupWebView()
-        }else{
+        } else {
             setResult(Activity.RESULT_OK, resultIntent)
             finish()
         }
--- a/sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
+++ b/sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
@@ -265,7 +265,7 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
         val resultIntent = Intent().apply {
             putExtra(EXTRA_AUTHORIZING_PAYMENT_RESULT, Failure(throwable))
         }
-        if (arrayOf("Challenge protocol error","Challenge runtime error","3DS2 initialization failed").contains(throwable.message)) {
+        if (arrayOf("Challenge protocol error", "Challenge runtime error", "3DS2 initialization failed").contains(throwable.message)) {
             setupWebView()
         } else {
             setResult(Activity.RESULT_OK, resultIntent)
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -265,8 +265,12 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
...
             putExtra(EXTRA_AUTHORIZING_PAYMENT_RESULT, Failure(throwable))
         }
-        setResult(Activity.RESULT_OK, resultIntent)
-        finish()
+        if(throwable.message == "3DS2 initialization failed"){
+            setupWebView()
+        }else{
+            setResult(Activity.RESULT_OK, resultIntent)
+            finish()
+        }
```
aashishgurung, line 273: Let's add spacing around `if` and `else`. In other places, `if` and `else` has spacing.

danfowler: @aashishgurung @AnasNaouchi can we set a standard formatter for Java?

aashishgurung: I found [checkStyle](https://github.com/checkstyle/checkstyle)

danfowler: Great. Thanks @aashishgurung

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -265,8 +265,12 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
         val resultIntent = Intent().apply {
             putExtra(EXTRA_AUTHORIZING_PAYMENT_RESULT, Failure(throwable))
         }
-        setResult(Activity.RESULT_OK, resultIntent)
-        finish()
+        if (arrayOf("Challenge protocol error","Challenge runtime error","3DS2 initialization failed").contains(throwable.message)) {
```
danfowler, line 268: Is there any better way of identifying the error other than this string error message?

vault087 (approved):
LGTM, besides the concern about using raw string identifiers instead of constants

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
AnasNaouchi: Currently the errors are passed to our custom exception class `OmiseException` as plain strings which is doing a compare may result in typos I would say.

danfowler: Should we create new custom exceptions for these cases?

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -265,8 +265,12 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
         val resultIntent = Intent().apply {
```
danfowler, line 265: On the above line, why `private fun finishActivityWithFailure(throwable: Throwable)` and not something like `private fun finishActivityWithFailure(e: Throwable)`

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
AnasNaouchi: @danfowler  I would say something like this:
```
enum class ErrorMessage(val message: String) {
    CHALLENGE_PROTOCOL_ERROR("Challenge protocol error"),
    CHALLENGE_RUNTIME_ERROR("Challenge runtime error"),
    INITIALIZATION_FAILED("3DS2 initialization failed")
}
```

danfowler: @AnasNaouchi OK, as agreed we'll make a new ticket and separate PR.

danfowler, line 268:
```suggestion
        if (arrayOf("Challenge protocol error", "Challenge runtime error", "3DS2 initialization failed").contains(throwable.message)) {
```

danfowler: @AnasNaouchi we need linting rules, but until we do, can we make sure to match the existing style?

PR #306, opened 2024-01-10
Upgrade targetSDKLevel and compileSDKLevel to 34

AnasNaouchi:
## Description

Upgrade the target sdk and the compile sdk to level 34 as a part of the overall upgrade initiatives.

*NOTE:*

The upgrade was performed on the POC branch as in master we faced some problems:
- Android SDK level 34 required that certain edits for any usage of `PendingIntent` must be performed but this code is not used in the SDK directly but it is part of another library `co.omise:threeds-android` that is no longer used in the SDK but is still a dependency as it is required in some parts of the app like cutom UI in the SDK. Removal of the library was not successful as it is needed for custom UI implementation by the merchant.
- Another attempt was to actually edit the code of the library but complications arose preventing local usage of the library and even if that did not happen we will have to release a new version of a library that is no longer used just to solve this problem.

For that the edits were performed in POC

### DIFF
--- a/sdk/src/main/java/co/omise/android/ui/CreditCardEditText.kt
+++ b/sdk/src/main/java/co/omise/android/ui/CreditCardEditText.kt
@@ -85,12 +85,12 @@ class CreditCardEditText : OmiseEditText {
         cardBrandImagePaint = Paint(Paint.ANTI_ALIAS_FLAG)
     }
 
-    override fun onDraw(canvas: Canvas?) {
+    override fun onDraw(canvas: Canvas) {
         super.onDraw(canvas)
         cardBrandImage?.let {
             val imageLeftPosition = width.toFloat() - it.width - paddingRight
             val imageTopPosition = (height - it.height) / 2f
-            canvas?.drawBitmap(it, imageLeftPosition, imageTopPosition, cardBrandImagePaint)
+            canvas.drawBitmap(it, imageLeftPosition, imageTopPosition, cardBrandImagePaint)
         }
     }
 
--- a/testsupport/build.gradle
+++ b/testsupport/build.gradle
@@ -2,12 +2,12 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 android {
-    compileSdkVersion 29
+    compileSdkVersion compile_sdk_version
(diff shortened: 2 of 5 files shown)


sdk/src/main/java/co/omise/android/ui/CreditCardEditText.kt
```diff
@@ -85,12 +85,12 @@ class CreditCardEditText : OmiseEditText {
         cardBrandImagePaint = Paint(Paint.ANTI_ALIAS_FLAG)
     }
 
-    override fun onDraw(canvas: Canvas?) {
+    override fun onDraw(canvas: Canvas) {
```
AnasNaouchi, line 88: The `nullable` method is not available anymore in api level 34

aashishgurung: Any idea why was it was initially designed as nullable? Will this change impacts our system negatively?

AnasNaouchi: The nullibility of a parameter may change due to many reasons but some of them are to have enhanced design patterns or to have a better usage of kotlin null safe features. 

The sdk is compiled separately from the settings of the host app so it should not affect the users. I have already tried it with the lowest allowed api on Google playstore (31) and it works fine.

aashishgurung (approved):

testsupport/build.gradle
```diff
@@ -2,12 +2,12 @@ apply plugin: 'com.android.library'
...
 apply plugin: 'kotlin-android-extensions'
 android {
-    compileSdkVersion 29
+    compileSdkVersion compile_sdk_version
 
 
     defaultConfig {
-        minSdkVersion 16
-        targetSdkVersion 29
+        minSdkVersion min_sdk_version
```
danfowler, line 9: @AnasNaouchi is it OK that the `minSdkVersion` here is going from 16 → 21?

AnasNaouchi: Yeah should not be a problem

danfowler (approved):

danfowler:
@AnasNaouchi Shall we update the PR description for https://github.com/omise/omise-android/pull/276 as we include things that are a little out of scope?

PR #311, opened 2024-01-30
Upgrade Android Gradle Plugin (AGP) 8.2.2

AnasNaouchi:
## Description

Upgrade to AGP 8.2.2 and remove deprecations where possible.

**Note:**
Upgrade to the latest version of android studio to be able to run the sdk/example app (Android Studio Hedgehog | 2023.1.1 Patch 2 )


danfowler:
@AnasNaouchi Probably worth pointing out that potential reviewers might want to update to the latest version of Android Studio as the version released in September only supports up to 8.1.2

app/src/kotlin/java/co/omise/android/example/CheckoutActivity.kt
```diff
@@ -65,11 +77,45 @@ class CheckoutActivity : AppCompatActivity() {
 
     private var capability: Capability? = null
 
+    private lateinit var  authorizingPaymentLauncher: ActivityResultLauncher<Intent>
```
danfowler, line 80:
```suggestion
    private lateinit var authorizingPaymentLauncher: ActivityResultLauncher<Intent>
```

app/src/java/java/co/omise/android/example/CheckoutActivity.java
```diff
@@ -12,8 +12,11 @@
 
 import org.jetbrains.annotations.NotNull;
 
+import androidx.activity.result.ActivityResultLauncher;
+import androidx.activity.result.contract.ActivityResultContracts;
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AppCompatActivity;
+
```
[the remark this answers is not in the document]
AnasNaouchi: separated all the `co.omise.android` imports from the other android imports

app/src/java/java/co/omise/android/example/CheckoutActivity.java
```diff
@@ -101,7 +107,7 @@ public void onRequestSucceed(@NotNull Capability model) {
 
             @Override
             public void onRequestFailed(@NotNull Throwable throwable) {
-                snackbar.setText(StringsKt.capitalize(throwable.getMessage())).show();
+                snackbar.setText(Objects.requireNonNull(capitalize(throwable.getMessage()))).show();
```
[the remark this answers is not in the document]
AnasNaouchi: ref: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/capitalize.html
The problem is that `throwable.getMessage()` itself can return `null`. So if we change our capitalize function to only accept non null values we will have to wrap `throwable.getMessage()` in require not null instead of wrapping the entire `capitalize` function. Either way we will have to put `requireNotNull` on one of the functions.
- Example
New code:
```java
    @NonNull
    public static String capitalize(@NotNull String input) {
        if (input.isEmpty()) {
            return input;
        }

        char firstChar = input.charAt(0);
        char capitalizedFirstChar = Character.toUpperCase(firstChar);

        return capitalizedFirstChar + input.substring(1);
    }
snackbar.setText(capitalize(Objects.requireNonNull(throwable.getMessage()))).show();
```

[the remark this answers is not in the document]
danfowler: Great, thanks

sdk/src/main/java/co/omise/android/models/ModelTypeResolver.kt
```diff
@@ -29,7 +29,7 @@ class ModelTypeResolver : TypeIdResolverBase() {
             (types as HashMap<String, Class<*>>)["document"] = Document::class.java
             (types as HashMap<String, Class<*>>)["authentication"] = Authentication::class.java
         }
-        return Collections.unmodifiableMap(types)
+        return Collections.unmodifiableMap(types!! as MutableMap<out String, out Class<*>>)
```
danfowler, line 32: @AnasNaouchi can we adjust to avoid this "force unwrap" of `types`? We have a `null` check above.

danfowler: @AnasNaouchi does this work?

```
    fun getKnownTypes(): Map<String, Class<*>> {
        val o = types as MutableMap<String, Class<*>>

        if (o == null) {
            o["card"] = Card::class.java
            o["token"] = Token::class.java
            o["capability"] = Capability::class.java
            o["payment_method"] = PaymentMethod::class.java
            o["source"] = Source::class.java
            o["error"] = APIError::class.java
            o["barcode"] = Barcode::class.java
            o["document"] = Document::class.java
            o["authentication"] = Authentication::class.java
        }
        return Collections.unmodifiableMap(o)
    }
```

AnasNaouchi: `Condition 'o == null' is always 'false'` so its not reachable

vault087 (approved):
LGTM

danfowler (approved):

PR #316, opened 2024-03-03
Add 3ds config endpoint

AnasNaouchi:
## Description

Replace hard coded Netcetera config values with endpoint values

### DIFF
--- a/sdk/src/main/java/co/omise/android/ThreeDS2ServiceWrapper.kt
+++ b/sdk/src/main/java/co/omise/android/ThreeDS2ServiceWrapper.kt
@@ -13,11 +13,7 @@ import com.netcetera.threeds.sdk.api.transaction.Transaction
 import com.netcetera.threeds.sdk.api.transaction.challenge.ChallengeParameters
 import com.netcetera.threeds.sdk.api.transaction.challenge.ChallengeStatusReceiver
 import com.netcetera.threeds.sdk.api.ui.logic.UiCustomization
-import java.security.MessageDigest
 import java.util.Collections
-import javax.crypto.Cipher
-import javax.crypto.spec.IvParameterSpec
-import javax.crypto.spec.SecretKeySpec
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/co/omise/android/ThreeDS2ServiceWrapper.kt
```diff
@@ -26,21 +32,56 @@ internal class ThreeDS2ServiceWrapper(
...
-    suspend fun initialize() =
+    // Computes SHA-512 hash of a string
+    fun hash512(data: String): ByteArray {
+        val hasher = MessageDigest.getInstance("SHA-512")
+        return hasher.digest(data.toByteArray(Charsets.UTF_8))
+    }
+
+    // Decrypts the encrypted data using the decryption key and AES/CTR/NoPadding cipher
+    fun aesDecrypt(
+        ciphertext: ByteArray,
```
aashishgurung, line 43: Should we consider creating a dedicated class for `hash512` and `aesDecrypt`?

vault087 (approved):

sdk/src/main/java/co/omise/android/ThreeDS2ServiceWrapper.kt
AnasNaouchi: If we were also encrypting stuff i would say we should consider that, but the usage is pretty limited and centered around 3DS.

aashishgurung: Although the usage is limited at the moment, we now have 2 reasons to update the class which violates SRP. One related to encryption and another related to 3DS.

Separation ensures that the primary class remains focused on its core responsibility of handling 3DS-specific logic, resulting in improved clarity and better comply with SRP.

aashishgurung (approved):

PR #320, opened 2024-03-08
Add acs check

AnasNaouchi:
## Description

Add a condition to check if the auth url requires 3DS processing. If not open web view.


vault087 (approved):

aashishgurung (approved):

pynkittisak:
Can we add unit test for that function?

AnasNaouchi:
> Can we add unit test for that function?

The thing is that this function is in a class that is covered by android instrumentation tests. These tests were so flaky so we disabled them and we have a task to fix. We will add once we fix the issue.

PR #323, opened 2024-03-20
Update Auth and Config models

AnasNaouchi:
## Description

Replace hard coded `messageVersion` and `acsReferenceNumber` with values from backend.

### DIFF
--- a/sdk/src/main/java/co/omise/android/models/NetceteraConfig.kt
+++ b/sdk/src/main/java/co/omise/android/models/NetceteraConfig.kt
@@ -1,4 +1,3 @@
-import co.omise.android.BuildConfig
 import co.omise.android.api.RequestBuilder
 import co.omise.android.models.Model
 import com.fasterxml.jackson.annotation.JsonProperty
--- a/sdk/src/main/java/co/omise/android/models/NetceteraConfig.kt
+++ b/sdk/src/main/java/co/omise/android/models/NetceteraConfig.kt
@@ -15,7 +15,7 @@ internal data class NetceteraConfig(
     @field:JsonProperty("device_info_encryption_cert_pem") val deviceInfoEncryptionCertPem: String? = null,
     @field:JsonProperty("directory_server_id") val directoryServerId: String? = null,
     @field:JsonProperty("key") val key: String? = null,
-    @field:JsonProperty("message_version") val messageVersion: String = BuildConfig.MESSAGE_VERSION,
+    @field:JsonProperty("messageVersion") val messageVersion: String? = null,
     override val modelObject: String? = null,
     override val id: String? = null,
     override val livemode: Boolean = false,
--- a/sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentViewModel.kt
+++ b/sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentViewModel.kt
@@ -174,7 +174,6 @@ internal class AuthorizingPaymentViewModel(
                 set3DSServerTransactionID(ares.threeDSServerTransID)
                 threeDSRequestorAppURL = createThreeDSRequestorAppURL(ares.sdkTransID)
                 acsTransactionID = ares.acsTransID
-                // TODO : check if where to get the sdkReferenceNumber value
                 acsRefNumber = ares.acsReferenceNumber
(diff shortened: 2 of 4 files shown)


sdk/src/main/java/co/omise/android/models/NetceteraConfig.kt
```diff
@@ -15,7 +14,7 @@ internal data class NetceteraConfig(
     @field:JsonProperty("device_info_encryption_cert_pem") val deviceInfoEncryptionCertPem: String? = null,
     @field:JsonProperty("directory_server_id") val directoryServerId: String? = null,
     @field:JsonProperty("key") val key: String? = null,
-    @field:JsonProperty("message_version") val messageVersion: String = BuildConfig.MESSAGE_VERSION,
+    @field:JsonProperty("messageVersion") val messageVersion: String? = null,
```
aashishgurung, line 17: Why is it `null` by default?

AnasNaouchi: we have to provide default values other wise we will have this error:
```kotlin
.InvalidDefinitionException: Cannot construct instance of `NetceteraConfig` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
```
 I think we will have to use [this](https://mvnrepository.com/artifact/com.fasterxml.jackson.module/jackson-module-kotlin) to fix it but for now this is how it is done.

aashishgurung (approved):

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentViewModel.kt
```diff
@@ -117,7 +116,7 @@ internal class AuthorizingPaymentViewModel(
     }
 
     private suspend fun sendAuthenticationRequest(netceteraConfig: NetceteraConfig) {
-        val transaction = threeDS2Service.createTransaction(netceteraConfig.directoryServerId!!, netceteraConfig.messageVersion)
+        val transaction = threeDS2Service.createTransaction(netceteraConfig.directoryServerId!!, netceteraConfig.messageVersion!!)
```
vault087, line 119: Does "!!" cause crash if value is NULL? If so it's always recommended to avoid it as some future changes could break it

vault087 (approved):

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentViewModel.kt
AnasNaouchi: It will throw an error that the param is null if it is missing from the response then it will open webview which is the expected behavior.

annatotest: Emvco recommends to use the latest 3ds protocol version if Message version is null

annatotest: image (69)

AnasNaouchi: but what if our MS does not support the latest version, then it will fail. It is not safe to leave it to Netcetera defaults.

AnasNaouchi: @annatotest Any other concerns ?

sdk/src/test/java/co/omise/android/ui/AuthorizingPaymentViewModelTest.kt
```diff
@@ -60,6 +60,7 @@ class AuthorizingPaymentViewModelTest {
             """-----BEGIN CERTIFICATE-----\r\nMIIDbzCCAlegAwIBAgIJANp1aztd\r\n-----END CERTIFICATE-----""",
             "TestId",
             "N2OZnlR1fDizs2SO5ukLnmaCupBaHcA=",
+            "2.2.0",
```
annatotest, line 63: Will we test with other versions?

AnasNaouchi: its just a demo number. It does not matter what we put here.

annatotest (approved):

PR #324, opened 2024-03-27
Update readme

AnasNaouchi:
## Description

Update README.md to include changes required to use the new 3DS functionalities and describe any new recommended practices.


README.md
```diff
@@ -27,13 +27,18 @@ Attestation of Compliance (AoC) delivered by a certified QSA Auditor.
 This SDK provides the means to tokenize card data on end-user mobile phone without the data
 having to go through your server.
 
+## Notice
+
+The minimum supported version of the SDK is 4.3.1. Please do not use any versions below that number as you will
```
aashishgurung, line 32: Let's highlight the version.

README.md
```diff
@@ -463,10 +520,11 @@ style.xml
...
-Some payment methods require the customer to authorize the payment using an authorization URL. This includes [3-D Secure verification](https://docs.opn.ooo/fraud-protection#3-d-secure), [Internet Banking payment](https://docs.opn.ooo/internet-banking), [Alipay](https://docs.opn.ooo/alipay), etc. Opn Payments Android SDK provides a built in class to handle the authorization.
 
-On payment methods that require opening the external app (e.g. mobile banking app) to authorize the transaction, set the *return_uri* to a **deeplink** or **applink** to be able to open the merchant app. Else, after the card holder completes authorizing the transaction on the external app, the flow redirects to the normal link in the *return_uri* and opens it on the browser app, and therefore results in the payment not being completed.
+Some payment methods require the customer to authorize the payment using an authorization URL. This includes [3-D Secure verification](https://docs.opn.ooo/fraud-protection#3-d-secure), [Internet Banking payment](https://docs.opn.ooo/internet-banking), [Alipay](https://docs.opn.ooo/alipay), etc. Opn Payments Android SDK provides a built in class to handle the authorization.
```
annatotest, line 524: Alipay is somewhat deprecated, in favor of Alipay + and others. Can we please swap it with [Mobile Banking SCB](https://docs.opn.ooo/mobile-banking-scb)?

annatotest: also 'built-in class' rather than built in class, I believe.

README.md
```diff
@@ -463,10 +520,11 @@ style.xml
...
 
+On payment methods that require opening the external app (e.g. mobile banking app) to authorize the transaction, set the _return_uri_ to a **deeplink** or **applink** to be able to open the merchant app. Else, after the card holder completes authorizing the transaction on the external app, the flow redirects to the normal link in the _return_uri_ and opens it on the browser app, and therefore results in the payment not being completed.
```
annatotest, line 526: I think instead of `Else, after the....` we better use `Otherwise after card holder completes`

annatotest: And could you please explain to me why payment will not be completed if I open return_uri in a browser?

README.md
```diff
@@ -482,22 +540,24 @@ file as follows:
...
-            startActivityForResult(this, AUTHORIZING_PAYMENT_REQUEST_CODE)
+            authorizingPaymentLauncher.launch(this)
         }
 }
 ```
 
-Replace the string `AUTHORIZED_URL` with the authorized URL that comes with the created charge and the array of string `EXPECTED_URL_PATTERNS` with the expected pattern of redirected URLs array.
-
+Replace the string `EXTRA_AUTHORIZED_URLSTRING` with the authorized URL that comes with the created charge and the array of string `EXTRA_EXPECTED_RETURN_URLSTRING_PATTERNS` with the expected pattern of redirected URLs array.
+The `EXTRA_UI_CUSTOMIZATION` parameter is used to customize the UI in the built in 3DS in app flow in the SDK and during the challenge flow.
```
annatotest, line 558: built-in

annatotest: better to say: The `EXTRA_UI_CUSTOMIZATION` parameter is used to customize UI of the built-in 3DS SDK during the 3DS challenge flow.

README.md
```diff
@@ -507,34 +567,82 @@ val toolbarCustomization = ToolbarCustomizationBuilder()
...
             ))
             .build()
 ```
 
-After the end-user completes the authorizing payment process, the activity result
-callback will be called. Handle it in this manner:
+You can check out the [UiCustomization](/sdk/src/main/java/co/omise/android/config/UiCustomization.kt) class to see customizable UI elements in the challenge flow.
+
+After the end-user completes the payment authorization process, the activity result
+callback will be and you will receive different responses based on how the transaction was processed
```
annatotest, line 581: I think we missed 'callback will be **called**' from the original sentence here. Also I think need to add **that** into the 'based on how the transaction **that** was processed'

README.md
```diff
@@ -507,34 +567,82 @@ val toolbarCustomization = ToolbarCustomizationBuilder()
...
+```kotlin
+val client = Client("pkey_test_1234")
+client.observeTokenUntilChargeStatusChanged("tokn_test_1234", object: RequestListener<Token> {
+    override fun onRequestSucceed(model: Token) {
+        TODO("Not yet implemented")
+    }
 
-### 3D Secure 2
+    override fun onRequestFailed(throwable: Throwable) {
+        TODO("Not yet implemented")
```
annatotest, line 638: Are we going to leave it like this? 
`TODO("Not yet implemented")`

README.md
```diff
@@ -507,34 +567,82 @@ val toolbarCustomization = ToolbarCustomizationBuilder()
...
+    }
+})
+```
 
-To support 3D Secure 2, you can check out the [3D Secure guide](docs/3d-secure-v2.md).
+### Authorizing payment via an external app
+
+Some request methods allow the user to authorize the payment with an external app, for example Alipay. When a user needs to authorize the payment with an external app, `AuthorizingPaymentActivity` will automatically open an external app. However merchant developers must handle the `Intent` callback by themselves.
```
annatotest, line 645: I think better do not use article here:
 'When user needs'

muthuswamyopn:
Do you want to make it "When users need", because "When user need" is wrong?

README.md
AnasNaouchi: @annatotest  The previous explanation was like this but i believe that the actual payment is completed but the callback to the app will not be completed as the browser will stay opened in the merchant's app so in order for the browser to close automatically the merchant needs to implement the described set up.

README.md
AnasNaouchi: @annatotest Yes the merchant will handle this on his side like show an error message for the user.

aashishgurung (approved):

PR #325, opened 2024-04-01
Handle config endpoint error and unavailability

AnasNaouchi:
## Description

Handle config endpoint errors and switch to webview when encountering an error .

### DIFF
--- a/sdk/src/main/java/co/omise/android/ThreeDSConfigProvider.kt
+++ b/sdk/src/main/java/co/omise/android/ThreeDSConfigProvider.kt
@@ -1,0 +1,33 @@
+import android.net.Uri
+import co.omise.android.AuthorizingPaymentURLVerifier
+import co.omise.android.api.Client
+import com.netcetera.threeds.sdk.api.exceptions.InvalidInputException
+import java.net.URL
+
+internal class ThreeDSConfigProvider(private val urlVerifier: AuthorizingPaymentURLVerifier, private val client: Client) {
+    // Use the authorize_uri to create the config url
+    fun createThreeDSConfigUrl(authUrl: String): String {
+        try {
+            // Check if the authUrl is valid
+            URL(authUrl)
+            val base = Uri.parse(authUrl).buildUpon().clearQuery().build()
+            val parts = base.toString().split('/').toMutableList()
+            parts[parts.lastIndex] = "config"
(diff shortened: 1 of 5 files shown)


pynkittisak (approved):

sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentViewModel.kt
```diff
@@ -115,6 +115,21 @@ internal class AuthorizingPaymentViewModel(
         }
     }
 
+    private suspend fun getConfigs(): NetceteraConfig? {
```
aashishgurung, line 118: Let's move `getConfigs` and createNetceteraConfigUrl` to a separate config class. ViewModel just needs to hold the data for the UI. It doesn't need to know where the data comes from.

AnasNaouchi: but we also have the `sendAuthenticationRequest` function that gets data for authentication also. That is why i kept the other functions.

aashishgurung: I understand it but if we continue doing this then we will keep on adding more code that should be here. There might be more stuff here that shouldn't be here but we can start by focusing on Netcetra config.

aashishgurung (approved):

pynkittisak (approved):

sdk/src/main/java/co/omise/android/ThreeDSConfigProvider.kt
```diff
@@ -0,0 +1,33 @@
...
+import co.omise.android.api.Client
+import com.netcetera.threeds.sdk.api.exceptions.InvalidInputException
+import java.net.URL
+
+internal class ThreeDSConfigProvider(private val urlVerifier: AuthorizingPaymentURLVerifier, private val client: Client) {
+    // Use the authorize_uri to create the config url
+    fun createThreeDSConfigUrl(authUrl: String): String {
+        try {
+            // Check if the authUrl is valid
+            URL(authUrl)
```
annatotest, line 12: Can we maybe change `authUrl` to `authorizeUri` as per Omise API? Because `authUrl` could be interpreted differently, for example, as Authentication Url or Authorization Url since both of them start with `Auth`.
image

PR #328, opened 2024-04-18
Add installment terms calculations

AnasNaouchi:
## Description

This PR aims to add the proper calculations for installments based on the installments guides in [OPN docs](https://docs.opn.ooo/installment-payments). The calculations are based on hard coded values that do not change frequently. 
The affected sections are the list of bank installments and the selection of terms for installments. 
Example:
Screenshot_1713411728

### DIFF
--- a/sdk/src/main/java/co/omise/android/ui/InstallmentTermChooserFragment.kt
+++ b/sdk/src/main/java/co/omise/android/ui/InstallmentTermChooserFragment.kt
@@ -32,31 +32,32 @@ internal class InstallmentTermChooserFragment : OmiseListFragment<InstallmentTer
         setHasOptionsMenu(true)
     }
 
-
     override fun listItems(): List<InstallmentTermResource> {
         val currency = requester!!.currency
-        val minimumInstallmentAmountPerType = mapOf(
-            SourceType.Installment.Bay to Amount.fromLocalAmount(500.0,currency),
-            SourceType.Installment.FirstChoice to Amount.fromLocalAmount(300.0,currency),
-            SourceType.Installment.Bbl to Amount.fromLocalAmount(500.0,currency),
-            SourceType.Installment.Mbb to Amount.fromLocalAmount(83.33,currency),
-            SourceType.Installment.Ktc to Amount.fromLocalAmount(300.0,currency),
-            SourceType.Installment.KBank to Amount.fromLocalAmount(300.0,currency),
-            SourceType.Installment.Scb to Amount.fromLocalAmount(500.0,currency),
-            SourceType.Installment.Ttb to Amount.fromLocalAmount(500.0,currency),
-            SourceType.Installment.Uob to Amount.fromLocalAmount(500.0,currency)
-        )
-        val interestRatePerType = mapOf(
-            SourceType.Installment.Bay to 0.0074,
-            SourceType.Installment.FirstChoice to 0.0116,
-            SourceType.Installment.Bbl to 0.0074,
-            SourceType.Installment.Mbb to 0.0,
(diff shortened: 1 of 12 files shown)


vault087 (dismissed):

sdk/src/main/java/co/omise/android/ui/InstallmentTermChooserFragment.kt
```diff
@@ -28,20 +33,58 @@ internal class InstallmentTermChooserFragment : OmiseListFragment<InstallmentTer
...
+                val sourceType = (installment?.backendType as? BackendType.Source)?.sourceType
+                val minimumAmount = minimumInstallmentAmountPerType[sourceType]
+                val req = requester
+                val amount = req!!.amount
+                var interestAmount = 0.0
+                if (zeroInterestInstallments == false) {
+                    val rate = interestRatePerType[sourceType] ?: 0.0
+                    interestAmount = amount.toDouble() * rate
+                }
+                val installmentAmountPerMonth = (amount / term) + interestAmount
```
aashishgurung, line 74: Since interest rate is calculated from the total, it needs to be added to the amount before dividing by terms.

```suggestion
                val installmentAmountPerMonth = (amount + interestAmount) / term
```

sdk/src/main/java/co/omise/android/ui/InstallmentTermChooserFragment.kt
```diff
@@ -63,13 +106,16 @@ internal class InstallmentTermChooserFragment : OmiseListFragment<InstallmentTer
...
-        fun newInstance(installment: PaymentMethod) =
-            InstallmentTermChooserFragment().apply {
-                arguments =
-                    Bundle().apply {
-                        putParcelable(EXTRA_INSTALLMENT, installment)
-                    }
-            }
+        fun newInstance(
+            installment: PaymentMethod,
+            zeroInterestInstallments: Boolean,
```
aashishgurung, line 113: Is `zeroInterestInstallments` needed here?

sdk/src/main/java/co/omise/android/ui/PaymentCreatorActivity.kt
```diff
@@ -260,15 +262,15 @@ private class PaymentCreatorNavigationImpl(
 
     override fun navigateToInstallmentChooser(allowedInstalls: List<PaymentMethod>) {
         val fragment =
-            InstallmentChooserFragment.newInstance(allowedInstalls).apply {
+            InstallmentChooserFragment.newInstance(allowedInstalls, amount, capability.limits?.installmentAmount?.min ?: 0).apply {
```
aashishgurung, line 265: IMO, it would improve readability if we assign `capability.limits?.installmentAmount?.min ?: 0)` to a variable and use it in the function signature instead.

AnasNaouchi: After some investigation it would not be very neat, as we will have to pass some stuff twice to propagate the value as intended if i want to implement you r suggestion in all the similar functions .. Better not do this now.

AnasNaouchi: maybe just store the value in a variable inside the function

sdk/src/main/java/co/omise/android/ui/InstallmentTermChooserFragment.kt
AnasNaouchi: I just noticed that i can access it from somewhere else ...

aashishgurung (approved):

vault087 (approved):

PR #348, opened 2024-06-11
Load intallment terms from capabilities

AnasNaouchi:
## Description

Enhancement:
- Load the capability object from the SDK side and the merchant is no longer required to load in the app using the SDK
- Hard coded values like the `installment_terms` and some lists of banks have been removed and are now loaded from capabilities.
- Cross filter payment methods between capabilities and merchant request, ex:
merchant requests methods 1,2,3
capability contains methods 2,3
result: 2,3 are displayed

### DIFF
--- a/sdk/src/main/java/co/omise/android/ui/PaymentCreatorActivity.kt
+++ b/sdk/src/main/java/co/omise/android/ui/PaymentCreatorActivity.kt
@@ -226,14 +226,14 @@ class PaymentCreatorActivity : OmiseActivity() {
             val selectedTokenizationMethods = merchantPassedCapabilities.tokenizationMethods
             if (selectedPaymentMethods != null) {
                 filteredPaymentMethods =
-                    capability.paymentMethods!!.filter { capMethod ->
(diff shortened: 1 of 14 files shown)


sdk/src/main/java/co/omise/android/ui/PaymentCreatorActivity.kt
```diff
@@ -405,7 +503,7 @@ private class PaymentCreatorNavigationImpl(
 interface PaymentCreatorRequester<T : Model> {
     val amount: Long
     val currency: String
-    val capability: Capability
+    var capability: Capability
```
aashishgurung, line 506: Is this intentionally making capability mutable?

AnasNaouchi: it should not be mutable, was part of a different approach that i was trying. Will restore it.

aashishgurung (dismissed):

aashishgurung (approved):

pynkittisak (approved):

PR #351, opened 2024-07-12
Bump v5.2.0

AnasNaouchi:
## Description

Bump `v5.2.0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## V5.2.0
+* **Enhancement:** Force load capabilities from api and remove hard coded values.
+
 ## V5.1.0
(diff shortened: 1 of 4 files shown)


aashishgurung (approved):

muthuswamyopn (dismissed):
Use these edited sentences:

The SDK will automatically fetch your capabilities, but you can use this EXTRA value to pass custom payment methods.


Declare a `capability` variable as a `Capability` object and pass it as the value for the `OmiseActivity.EXTRA_CAPABILITY` key for your `Intent.` This way, you can pass custom payment methods to the payment methods list. The displayed payment methods will only be those supported by your capabilities that the SDK automatically fetches.


If you want to edit only the interest, you can create your capabilities as shown. When the SDK detects that you did not pass any payment methods, it will automatically use what is available.

   > You will see an empty list if your account supports none of the payment methods.

aashishgurung (dismissed):

timothyOmise (approved):

muthuswamyopn (commented):
If **none** of the payment methods are supported by your account you will see an empty list.

aashishgurung (approved):

gaurab13 (approved):

PR #355, opened 2024-09-08
Handle non existing deeplink app

AnasNaouchi:
## Description

Prevent the SDK from terminating the webview when a deep link fails to open like in a scenario of a bank app not installed and the user is trying to pay using a payment methods that needs that app. The browser will wait for further action from the user when the app is not installed.
If the app is installed the behavior remains unchanged.


sdk/src/main/java/co/omise/android/ui/AuthorizingPaymentActivity.kt
```diff
@@ -303,7 +306,9 @@ class AuthorizingPaymentActivity : AppCompatActivity() {
         data: Intent?,
     ) {
         if (requestCode == REQUEST_EXTERNAL_CODE) {
-            finishActivityWithSuccessful(data)
+            if (isDeepLinkSuccOpened) {
```
aashishgurung, line 309:
```suggestion
           if (requestCode == REQUEST_EXTERNAL_CODE && isDeepLinkSuccOpened) {
               finishActivityWithSuccessful(data)
           }
```

aashishgurung (approved):

mmkittisak (approved):

